PR #1, opened 2022-04-24
Source: Added Mangafox

pandeynmn:
Adding Mangafox to the community source repository.


src/as/en.mangafox/src/helper.ts
```diff
@@ -0,0 +1,189 @@
...
+            const id    = url.replace('/manga/', '').replace('/1.html', '');
+            const spl   = id.split('/');
+
+            let title = '';
+            const titleSplit = element.select('p.title3').text().trim().split('-');
+            if (titleSplit.length >= 2) title = titleSplit.slice(1).join('-');
+
+            let dateString = element.select('.title2').text().trim();
+            let dateObject = ValueRef.string(dateString);
+            let date = dateObject.toDate('MMM d,yyyy');
```
Skittyblock, line 75: This uses the user's current locale, but the site uses US locale regardless. As such, this should be
```ts
dateObject.toDate('MMM d,yyyy', 'en_US');
```

Skittyblock (changes requested):

Skittyblock (approved):

src/as/en.mangafox/src/helper.ts
pandeynmn: Implemented the change

PR #6, opened 2022-05-13
Add Mangstream template

daarkdemon:
-  Alphascans
  -  Asurascans
  -  Blackarmy
  -  Kumascans
  -  Readkomik


Skittyblock:
This has issues with parsing sources with different languages (e.g. they have different words for the status like "ongoing"), so if you can fix that I can merge it.

Skittyblock:
Well, it really _should_ be the source and not the template that defines the overrides for the different language strings, but this works for now so I guess it's okay. If more sources are added that use different language strings, this will need to change.

PR #7, opened 2022-05-15
Mangapill 

pandeynmn:
Added single page/listing source Mangapill


Skittyblock:
You should add DeepLink support. Manga links and chapter links are just the id you use appended to the base url, so it shouldn't be tough to parse. Just trim the base url off and pass the id to your parser functions.

Skittyblock (changes requested):
DeepLink support

Skittyblock (approved):

PR #20, opened 2022-05-25
Madara sources 

pandeynmn:
Madara sources idk


src/rust/madara/sources/coloredmanga/res/filters.json
```diff
@@ -0,0 +1,87 @@
+[
+	{ "type": "title" },
+	{
+		"type": "select",
+		"name": "Adult",
+		"options": ["All", "No Adult Content", "Only Adult Contnt"]
```
beerpiss, line 6:
```suggestion
		"options": ["All", "No Adult Content", "Only Adult Content"]
```

src/rust/madara/template/src/template.rs
```diff
@@ -0,0 +1,344 @@
...
+	if did_search {
+		get_search_result(data, url)
+	} else {
+		get_series_page(data, "_latest_update", page)
+	}
+}
+
+pub fn get_search_result(data: MadaraSiteData, url: String) -> Result<MangaPageResult> {
+	println!("get_search_result: {}", url);
+
```
Skittyblock, line 55: Don't keep unneeded logs.
```suggestion
```

src/rust/madara/template/src/template.rs
```diff
@@ -0,0 +1,344 @@
...
+	page: i32,
+) -> Result<MangaPageResult> {
+	if _listing.name == "Popular" {
+		return get_series_page(data, "_wp_manga_views", page);
+	}
+	if _listing.name == "Trending" {
+		return get_series_page(data, "_wp_manga_week_views_value", page);
+	}
+	return get_series_page(data, "_latest_update", page);
+}
```
Skittyblock, line 160: You're using this parameter, so it shouldn't be prefixed by an underscore.
```suggestion
pub fn get_manga_listing(
	data: MadaraSiteData,
	listing: Listing,
	page: i32,
) -> Result<MangaPageResult> {
	if listing.name == "Popular" {
		return get_series_page(data, "_wp_manga_views", page);
	}
	if listing.name == "Trending" {
		return get_series_page(data, "_wp_manga_week_views_value", page);
	}
	return get_series_page(data, "_latest_update", page);
}
```

src/rust/madara/template/src/template.rs
```diff
@@ -0,0 +1,344 @@
...
+	Ok(pages)
+}
+
+pub fn modify_image_request(base_url: String, request: Request) {
+	request.header("Referer", &base_url);
+}
+
+pub fn handle_url(url: String, data: MadaraSiteData) -> Result<DeepLink> {
+	Ok(DeepLink {
+		manga: Some(get_manga_details(url.clone(), data)?),
```
Skittyblock, line 341: Don't think this works. `get_manga_details` takes a manga id, not a url, so you need to parse out the manga id from the url. Also, you should consider how chapter urls are handled in addition to manga urls, even if you don't return the chapter.

Skittyblock (changes requested):

PR #27, opened 2022-06-08
new template: Otaku Sanctuary

beerpiss:
Includes [Otaku Sanctuary](https://otakusan.net) and [MyRockManga](https://myrockmanga.com). Both sources are fairly similar (like MangaSee vs MangaLife), but I took the template approach because these sources sometimes varies on what they host, mostly because of DMCA takedowns, ~~and also because they have different icons~~

Sources are primarily Vietnamese, though there is a significant number of English, Italian, French and Spanish mangas.


src/rust/otaku_sanctuary/build.sh
```diff
@@ -0,0 +1,37 @@
+# template source build script
+# usage: ./build.sh [source_name/-a]
+
+if [ "$1" != "-a" ]; then
```
Skittyblock, line 4: I'd like it so that `./build.sh` just builds all the sources too, rather than having to specify `-a`.
```suggestion
if [ "$1" != "-a" ] && [ "$1" != "" ]; then
```

Skittyblock (changes requested):
Couldn't really get the manga details pages to load myself, but that's because of my location so oh well. I'll trust that it works fine.

Also, I think you should change the `otaku_sanctuary` directory to `otakusan`, `otakusanctuary`, or `otaku-sanctuary` since the current directory names conform to a typical bundle id format.

Skittyblock (approved):

PR #33, opened 2022-06-10
Add support for readm.org

daarkdemon:


Skittyblock:
Remember to run the formatter (see commits for details) in the future before PRing so it saves me some time :)

PR #34, opened 2022-06-11
new sources: Manga-chan, Yaoi-chan, Hentai-chan

beerpiss:
[Manga-chan](https://manga-chan.me), [Yaoi-chan](https://yaoi-chan.me), [Hentai-chan](https://y.hentaichan.live) are Russian manga sources. A few strings were translated with DeepL since I don't know Russian myself.


src/rust/multichan/build.sh
```diff
@@ -0,0 +1,37 @@
+# template source build script
+# usage: ./build.sh [source_name/-a]
+
+if [ "$1" != "-a" ]; then
```
Skittyblock, line 4: :trollface: 
```suggestion
if [ "$1" != "-a" ] && [ "$1" != "" ]; then
```

src/rust/multichan/template/src/helper.rs
```diff
@@ -0,0 +1,83 @@
+use aidoku::{std::html::Node, std::String, std::Vec, MangaContentRating, MangaViewer};
```
Skittyblock, line 1:
```suggestion
use aidoku::{std::html::Node, std::String, std::Vec, MangaContentRating, MangaViewer};

```

Skittyblock (changes requested):

src/rust/multichan/template/src/template.rs
```diff
@@ -0,0 +1,391 @@
...
+				need_sort_date=if order_by_date_when_search { "&need_sort_date=true" } else { "" },
+			)
+		};
+		let html = Request::new(&url, HttpMethod::Get).html();
+		self.parse_manga_list(html, if !title.is_empty() { 40 } else { 20 })
+	}
+
+	pub fn get_manga_listing(&self, listing: Listing, _: i32) -> Result<MangaPageResult> {
+		if &listing.name == "Случайная" {
+			let html = Request::new("https://manga-chan.me/manga/random", HttpMethod::Get).html();
```
Skittyblock, line 176: I think this messes up the other two sources because they also have the same listing but you're only reading from the manga-chan website. Do the other sites have the same endpoint? You can probably just use the base_url here.

beerpiss: oh oops that was my bad, other sources have the same endpoints too

src/rust/multichan/template/src/template.rs
```diff
@@ -0,0 +1,392 @@
...
+		} else if let Some(begin_) = html.find("fullimg\": [") {
+			// hentai-chan
+			println!("hentaichan");
+			let begin = begin_ + 11;
+			let end = html[begin..].find(']').map(|i| i + begin).unwrap_or(0);
+			(begin, end)
+		} else {
+			(0, 2)
+		};
+		println!("{}", &html[begin..end]);
```
Skittyblock, line 324: Probably best not to have printing in public sources unless necessary.
```suggestion
		} else if let Some(begin_) = html.find("fullimg\": [") {
			// hentai-chan
			let begin = begin_ + 11;
			let end = html[begin..].find(']').map(|i| i + begin).unwrap_or(0);
			(begin, end)
		} else {
			(0, 2)
		};
```

Skittyblock (changes requested):

PR #43, opened 2022-06-21
[template] Add support for ninemanga(all languages)

daarkdemon:


beerpiss (changes requested):
Rename source folders to match their Cargo package names (e.g. `Ninemanga_br` -> `ninemanga_br`). It probably worked on your end since the Windows filesystem is case-insensitive.

daarkdemon:
Filters are different for each language. If we have dynamic filter generation then we can merge in one

PR #47, opened 2022-07-04
mangadex fixes

### LINKED ISSUE #25: MangaDex source doesn't display chapters incorrectly marked as having 0 pages
Example being [chapter 86 of The Ancient Magus' Bride](https://api.mangadex.org/manga/195023bf-cf9a-4772-94ef-09dd6eddea84/feed?order[volume]=desc&order[chapter]=desc&limit=500&contentRating[]=pornographic&contentRating[]=erotica&contentRating[]=suggestive&contentRating[]=safe&includes[]=scanlation_group&translatedLanguage[]=en). The reason the MangaDex source does this is because external chapters (which the source can't read) are marked as having 0 pages. I'm guessing that checking if the externalUrl parameter is null might be a better solution.

beerpiss:
Closes Aidoku/Aidoku#83
Closes #25 

Waiting on Aidoku/Aidoku#87

#### Features
* feat(filter): exclude original languages
* feat(filter): included/excluded tags mode
* feat(filter): has available chapters
* feat(settings): setting for blocking uploaders
* feat(settings): setting to force image servers to use HTTPS port 443
* feat(source): ratelimit requests with automatic retrying (3 requests per second)

#### Bug fixes
* fix(get_chapter_list): properly fetch all chapters for mangas with >=500 of them
* fix(get_chapter_list): deal with chapters incorrectly marked as having 0 pages
* fix(get_manga_listing): batch manga requests on the "Latest" listing to reduce load
* fix: pre-allocate vectors when possible to reduce allocation

#### Others
* chore(filter): split tags into 4 sections (contents, formats, themes, genres)
* chore: add more languages
* chore: replace match maps with filter ids
* chore: fix clippy lints
* chore: formatting

### Checklist


src/rust/multi.mangadex/src/parser.rs
```diff
@@ -1,39 +1,62 @@
...
+	// Try for preferred language first
+	match obj.get(&language).as_string() {
+		Ok(value) => value.read(),
+		// Fallback to English
+		Err(_) => match obj.get("en").as_string() {
+			Ok(value) => value.read(),
+			// Fallback to Japanese (might be romaji)
+			Err(_) => match obj.get("ja-ro").as_string() {
+				Ok(value) => value.read(),
+				// Fuck it, get first value
```
Skittyblock, line 26: family friendly 😃 
```suggestion
				// Screw it, get first value
```

src/rust/multi.mangadex/src/lib.rs
```diff
@@ -1,311 +1,282 @@
...
 #[get_manga_list]
 fn get_manga_list(filters: Vec<Filter>, page: i32) -> Result<MangaPageResult> {
 	let offset = (page - 1) * 20;
-	let mut url = String::from("https://api.mangadex.org/manga/?includes[]=cover_art&limit=20&offset=");
-	url.push_str(&i32_to_string(offset));
+	let mut url = String::from(
+		"https://api.mangadex.org/manga/?includes[]=cover_art\
+		&limit=20\
+		&offset=",
+	) + itoa::Buffer::new().format(offset);
```
Skittyblock, line 46: any reason to use itoa? I don't think i32_to_string was bad

Skittyblock (approved):
your aidoku-rs update wraps the rate limiting and new net functions doesn't it? is there any reason you just manually link them here instead of using the new update?

beerpiss:
because migrating the entire source to the rewrite is a pain since there are a lot of breaking changes

the old wrapper still works anyway

src/rust/multi.mangadex/src/lib.rs
beerpiss: It's [fast](https://docs.rs/itoa/latest/itoa/#performance-lower-is-better) and it doesn't take up that much space so I figured why not eh

PR #50, opened 2022-07-10
Add contribution guide and dual-license (Apache-2.0 OR MIT)

beerpiss:
### Description
- Add a contribution guide
- Dual license under either of Apache 2.0 or MIT. This is what many Rust projects do, so I figured this choice would be fitting. Need all contributors to be on board though.


beerpiss:
> Looks good. Is there any reason to actually dual-license with Apache 2 and MIT? I thought that both of them were compatible so they could be used under a project with either license regardless.

It's mostly the convention of Rust projects to adopt Apache-2.0 OR MIT since that's what the Rust language itself is licensed in.

PR #56, opened 2022-07-21
Fix NHentai user agent issue

phiefferj24:
Checklist:
none cuz idc


src/rust/multi.nhentai/src/lib.rs
```diff
@@ -83,7 +83,9 @@ fn get_manga_list(filters: Vec<Filter>, page: i32) -> Result<MangaPageResult> {
 
 	if is_sauce_code {
 		let url = helper::get_details_url(sauce_code);
-		let json = Request::new(&url, HttpMethod::Get).json().as_object()?;
+		let request = Request::new(&url, HttpMethod::Get);
+		let request = request.header("User-Agent", "Aidoku");
```
Skittyblock, line 87: you don't need to define the request variable twice just do
```suggestion
		let request = Request::new(&url, HttpMethod::Get)
			.header("User-Agent", "Aidoku");
```
for all of them to make it simple.

Skittyblock (changes requested):
also can you merge main into here so the workflow can run thanks

Skittyblock (approved):
alright probably fine

PR #65, opened 2022-07-29
Image url fix

daarkdemon:
Fix chapter image url containing whitespaces

Checklist:


src/rust/mangastream/template/src/template.rs
```diff
@@ -336,7 +336,7 @@ impl MangaStreamSource {
 		} else {
 			for (at, page) in html.select(self.page_selector).array().enumerate() {
 				let page_node = page.as_node();
-				let page_url = page_node.attr(self.page_url).read();
+				let page_url = page_node.attr(self.page_url).read().replace(' ', "%20");
```
beerpiss, line 339: why not `urlencode` the URL at this point, change the `urlencode` function to not encode `;,/?:@&=+$-_.!~*'()#` and you should be good

beerpiss (changes requested):
You should also bump the version of any sources affected by this issue.

src/rust/mangastream/template/src/template.rs
beerpiss:
```rust
pub fn urlencode<T: AsRef<[u8]>>(url: T) -> String {
    let bytes = url.as_ref();
    let hex = "0123456789ABCDEF".as_bytes();

    let mut result: Vec<u8> = Vec::with_capacity(bytes.len() * 3);

    for byte in bytes {
        let curr = *byte;
        if curr.is_ascii_alphanumeric() || b";,/?:@&=+$-_.!~*'()#".contains(&curr) {
            result.push(curr);
        } else {
            result.push(b'%');
            result.push(hex[curr as usize >> 4]);
            result.push(hex[curr as usize & 15]);
        }
    }
    String::from_utf8(result).unwrap_or_default()
}
```

daarkdemon: Made requested changes

beerpiss: you still haven't fixed the specific line I commented on lol

beerpiss (commented):
is it only kiryuu that is affected from the bug

src/rust/mangastream/template/src/template.rs
daarkdemon: I have changed function to url encode for the mentioned line 
Only kiryuu has been affected by this

daarkdemon:
> is it only kiryuu that is affected from the bug

Yes

beerpiss (approved):

PR #75, opened 2022-08-07
Changed source-path string

pandeynmn:
changed from manga to devmax
reset-scans.com/devmax/mangaid

Should not affect ids

Checklist:


beerpiss (approved):
deal with lints when you get around to a madara update

PR #77, opened 2022-08-16
Add MangAdventure template source

ObserverOfTime:
Checklist:


src/rust/mangadventure/template/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+		));
+		let results = get_value!(json, results, as_array);
+		let mut chapters = vec_from_array::<Chapter>(&results);
+
+		for result in results {
+			let obj = result.as_object().unwrap();
+			let mut url = get_value!(obj, url, as_string).read();
+			url.insert_str(0, self.base_url);
+			let title = get_value!(obj, full_title, as_string).read();
+			let chapter = get_value!(obj, number, as_float) as f32;
```
Skittyblock, line 234: This fails to parse sometimes (when chapter has a decimal point I think) but... I think this is an issue with the app? It's failing to detect the object as a float and aidoku-rs is failing `as_float()` since it makes sure the type is listed as a float, but the app returns the unknown type.

I fixed this by checking if the value could be converted to a NSNumber in the app, but that means an app update would be required before these sources work. If you wanted you could probably figure out a way to bypass the type checking and force it to convert to a float, otherwise I'll just have to hold off on merging for now.

Also there should probably be a -1.0 fallback value just in case.

src/rust/mangadventure/template/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+			self.base_url, id
+		));
+		let results = get_value!(json, results, as_array);
+		let mut chapters = vec_from_array::<Chapter>(&results);
+
+		for result in results {
+			let obj = result.as_object().unwrap();
+			let mut url = get_value!(obj, url, as_string).read();
+			url.insert_str(0, self.base_url);
+			let title = get_value!(obj, full_title, as_string).read();
```
Skittyblock, line 233: Wouldn't it make more sense to use the title key? full_title includes the volume and chapter number which Aidoku already displays, so it's duplicate information.
```suggestion
			let title = get_value!(obj, title, as_string).read();
```

Skittyblock (changes requested):

src/rust/mangadventure/template/src/lib.rs
Skittyblock: `as_float` works as expected with number types, but the app was providing the wrong type signature for the number which caused it not to work. https://github.com/Aidoku/Aidoku/commit/5d74c5c5f17c7c8d18b45601b91a3d8b8c708b0d fixes it without changes to the source, but it won't be in a new public build for a little while since I'm working on a lot of other stuff in the app before that can happen.

ObserverOfTime: Are licensed status and artist filter part of that other stuff? :smile:

Skittyblock: wdym by licensed status?

ObserverOfTime: Some series are officially licensed and don't have any chapters available on the site. Other extensions like MangaDex need that too.

Skittyblock: So are you saying you'd like the ability to display something like an explanation as to why chapters aren't available?

Skittyblock: I'll keep that in mind when I'm working on the source api overhaul. Unfortunately the api as it is right now is a little limited and doesn't allow for sending errors, etc. to the app.

Skittyblock (approved):
I think it should work fine now that 0.5 build 3 is out.

PR #97, opened 2022-11-13
fix(toonily): update search path and selector

KaBankz:
Checklist:


Skittyblock (changes requested):
search is happening in "family mode" and I'm guessing that's probably not ideal. I think "&adult" needs to be appended to the search url.

KaBankz:
> search is happening in "family mode" and I'm guessing that's probably not ideal. I think "&adult" needs to be appended to the search url.

Turns out toonily has a custom cookie (`toonily-mature=1`) that needs to be passed in for nsfw results to appear, it does not abide by the default `wpmanga-adault=1` cookie.

So, I updated the madara template to allow for passing in custom search cookies, and toonily search is working fine now...

Except, searching in toonily causes an infinite loop, and search results are duplicated infinitely, then the app crashes. The pagination for search requests is borked and even if there is only one page of results, it thinks there are more pages of results and requests them until the app crashes.

I tried looking around as to why this is happening, but I had no luck thus far.

Skittyblock:
Looks like toonily formats the search url a bit differently. Madara template defines it as `/search/{page}/?s={string}&...`, but toonily's is like `/search/{string}/page/{page}/?...`. However, it appears like it works the same if you use the s URL param still and just replace the path component with any string, so like `/search/x/page/{page}/?s={string}&...` works I think.

The template should probably be changed to allow for this, but I'm not sure if it's something unique to toonily or if other madara sources do this as well.

KaBankz:
You were right, the search url was messed up, I fixed toonily's search path but toonily search broke if `&post_type=wp-manga` was set. So I updated the template and helper to make `post_type` an overrideable variable, and everything works as intended now.

Skittyblock (commented):
Only Toonily version should be bumped since the functionality of the other sources has stayed the same. Otherwise looks good.

KaBankz:
> Only Toonily version should be bumped since the functionality of the other sources has stayed the same. Otherwise looks good.

Oh, I thought you're supposed to update all source versions for template updates.

No big deal tho, I'll revert that commit real quick

Skittyblock:
Typically yeah, but there's no reason to push updates out to everyone when nothing in the source was changed.

PR #99, opened 2022-11-15
feat(madara): add source lhtranslation

KaBankz:
Everything works fine except for loading certain older chapter images as the image URLs formats have changed.

The problem is with the URL encoding from the `img_url_encode()` helper.

https://github.com/Skittyblock/aidoku-community-sources/blob/de54c636ff5df0a0a4a1bc12a32c38324723d49d/src/rust/madara/template/src/helper.rs#L30-L63

It encodes the end of URLs for older images incorrectly, but newer images work fine.

For example:
Original: `c001---026-%255BLHTranslation%255D.jpg` ✅
Encoded: `c001---026-%5BLHTranslation%5D.jpg` ❌

I'm not sure how the encoding part works, so I'm clueless as to how to fix that issue 🤷‍♂️, but besides that, all is good.

Checklist:


Skittyblock:
I'm confused. `c001---026-%255BLHTranslation%255D.jpg` is the correct one while `c001---026-%5BLHTranslation%5D.jpg` is the result? I have a feeling it's the other way around. `%25` is literally `%` encoded, while `%5D` is `]`. I think it's trying to encode the `%` signs instead of just leaving them, so the solution would be to either just skip them (probably the easiest) or decode the image URL before encoding it (e.g. replace the `%5B` and `%5D` with `[` and `]`).

If you gave me a manga/chapter to look at I could double-check it.

KaBankz:
`Makikomarete Isekai Teni suru Yatsu wa, Taitei Cheat` any chapter between 1 - 34.2 is broken

Skittyblock:
Ok yep, the correct url is `.../c001---001-%5BLHTranslation%5D.jpg`. The easiest solution is probably just to modify `img_url_encode` like
```rs
...
 	result.push(b'-'); 
 } else if curr == b'%' { 
 	result.push(b'%'); 
 } else if curr == b'.' { 
 	result.push(b'.'); 
...
```

KaBankz:
Yeah, your fix worked, and everything is working fine now.

Skittyblock (approved):

PR #106, opened 2022-11-19
feat(madara): add setsuscans

KaBankz:
I'm not sure if I did everything the best way, but it worked.

Notable things that happened:
1. Update helper to not encode "?"
2. Update template to allow for setting a custom selector for the description
3. Update template to allow for setting a custom get_manga_id function (I tried, it worked)
4. Add base64 dependency to setsuscans and handle decoding its manga id

Checklist:


src/rust/madara/sources/setsuscans/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+// so this function overrides the default get_manga_id function to handle SetsuScans
+fn get_manga_id(manga_id: String, base_url: String, path: String) -> String {
+	let url = base_url + "/" + path.as_str() + "/" + manga_id.as_str();
+	let html = Request::new(url.as_str(), HttpMethod::Get).html();
+	let id_html = html.select("script#wp-manga-js-extra").attr("src").read();
+	let id_html = id_html.replace("data:text/javascript;base64,", "");
+	let decoded_html = base64::decode(id_html).unwrap();
+	let decoded_html = String::from_utf8(decoded_html).unwrap();
+	let id = &decoded_html
+		[decoded_html.find("manga_id").unwrap() + 11..decoded_html.find("\"}").unwrap()];
```
Skittyblock, line 30: It's not usually good to use `unwrap()`. Is this always going to exist for every manga? Is there any way to recover from a failure here (I'm guessing probably not)? It's better practice to do `expect()` so if it does end up failing in the future we can see why, rather than having to dig through the code too much.

Skittyblock:
Changes look fine to me I think. Only issue that might pop up is that the owner of SetsuScans requested it be removed from Tachiyomi, but I guess we can wait and see if that happens here or not since Aidoku is a lot smaller. It probably won't direct too much traffic away from the site.

src/rust/madara/sources/setsuscans/src/lib.rs
KaBankz: I used `expect()` at first, but changed to `unwrap()` because I wanted it to be consistent with the existing `get_int_manga_id` function. But, since it's bad practice, I'll switch it back to `expect()`.

Skittyblock (approved):

PR #109, opened 2022-11-21
feat(madara): add multi-language source leviatanscans

KaBankz:
leviatanscans is a multi-language source, and the madara template does not have any multi-language support as far as I could see; So, I just took the asurascans `get_lang_code` function and added it to the madara helper, and used it in leviatanscans to do some matching and change the url's and selectors when the user changes the source language. I don't know if this was the best implementation of multi-language support, but for now it works.

Checklist:


src/rust/madara/template/src/helper.rs
```diff
@@ -195,3 +195,13 @@ pub fn get_int_manga_id(manga_id: String, base_url: String, path: String) -> Str
 	let id = &id_html[id_html.find("manga_id").unwrap() + 11..id_html.find("\"};").unwrap()];
 	String::from(id)
 }
+
+pub fn get_lang_code() -> String {
```
Skittyblock, line 199: I think it'd be better to use an `Option<String>` here, so that way you don't have to use a mutable and can return None instead of another allocated string. Then, in the leviatan source, you can fall back to "en" as the default language.

src/rust/madara/template/src/template.rs
```diff
@@ -66,12 +67,15 @@ impl Default for MadaraSiteData {
 			post_type: String::from("wp-manga"),
 			// p to select description from
 			description_selector: String::from("div.description-summary div p"),
+			// selector for chapter list
+			chapter_selector: String::from("li.wp-manga-chapter  "),
```
Skittyblock, line 71: I know it had the spaces previously but I'm pretty sure they're not necessary.
```suggestion
			chapter_selector: String::from("li.wp-manga-chapter"),
```

Skittyblock (changes requested):
otherwise looks good I think

Skittyblock (approved):

PR #123, opened 2022-12-24
feat: add source zero scans

KaBankz:
Caveats:
- No Search - Zero Scans does not have a search api, instead they do client side json filtering (I tried implementing a fuzzy title match, but rust hates me)
- Loading chapter lists can be slow - Zero Scans paginates their chapters api, 20 per request so large chapter counts mean more requests, so longer wait
- To get the chapter list you need the manga id, right now the `manga_id` parameter contains the manga slug which is used for every other request except this one, this one needs the id. So an initial request is made solely to get the manga id, which adds more loading delay.

Checklist:


src/rust/en.zeroscans/src/helper.rs
```diff
@@ -0,0 +1,25 @@
...
+	match cleaned_time_ago
+		.to_uppercase()
+		.split_whitespace()
+		.last()
+		.unwrap_or("")
+	{
+		"YEAR" | "YEARS" => current_date() - (number * 31556926.0),
+		"MONTH" | "MONTHS" => current_date() - (number * 2629743.0),
+		"WEEK" | "WEEKS" => current_date() - (number * 604800.0),
+		"DAY" | "DAYS" => current_date() - (number * 86400.0),
```
Skittyblock, line 22: probably better to do the seconds per period multiplied out rather than constant numbers so it's clearer. e.g. `60 * 60 * 24 (=86400)` for 60s * 60m * 24hr = 86400s in a day, then add `* 7` for a week, etc.

Skittyblock:
Might be worth just filtering without the fuzzy search? You can do it case insensitive, but I think there are enough titles on the source to make it a pain just to use without having a search.

KaBankz:
> Might be worth just filtering without the fuzzy search?

The thing is without fuzzy search, you have to know the exact spelling of the manga you're looking for, which isn't the best user experience, but you're right, it's better than nothing, so I implemented a simple case insensitive title match.

Hopefully in the future a fuzzy search could be added, also looking at Asura Scans beta site, they also have a client side search so having an implementation of fuzzy matching could be useful in these cases. The [sublime_fuzzy](https://crates.io/crates/sublime_fuzzy) crate does not work because of dependency implementation issues, but maybe that could be fixed in the future.

Skittyblock (approved):

Skittyblock:
yeah sounds good

PR #127, opened 2022-12-27
fix(manhuaus): fix broken images for some series

KaBankz:
Checklist:

### DIFF
--- a/src/rust/madara/sources/manhuaus/res/source.json
+++ b/src/rust/madara/sources/manhuaus/res/source.json
@@ -3,7 +3,7 @@
 		"id": "en.manhuaus",
 		"lang": "en",
 		"name": "ManhuaUS",
-		"version": 1,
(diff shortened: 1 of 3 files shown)


Skittyblock (approved):
outside the scope of this pr, but I'm thinking we should probably have some of these madara sources default to scroll reading mode since they're all manhua/manhwa

KaBankz:
> outside the scope of this pr, but I'm thinking we should probably have some of these madara sources default to scroll reading mode since they're all manhua/manhwa

Definitely, lots of the madara sources have tags specifying what type of series it is (manga, manhwa, manhua). So it could be as simple as adding a check in template to look for these tags and set the reading mode accordingly. In fact this can also work for setting content rating, look for tags like adult, ecchi, mature, etc.

If that sounds good, I could pr it sometime this week.

PR #128, opened 2022-12-29
fix(mangadex): parse dates in correct timezone

KaBankz:
MangaDex uses UTC timestamps, without this fix new chapters would not show up for hours because the timestamps were parsed as future times.

Checklist:


Skittyblock:
oh so that was the issue lol

Skittyblock (approved):

PR #133, opened 2022-12-31
fix(mangastream)!: fix title parsing, use proper id's and update dependencies

KaBankz:
This fixes the repetitive chapter titles, "Ch. 123 - Chapter 123" and should be able to parse titles properly now.

> **Warning** **BREAKING CHANGE!**: Now proper id's are used, instead of full url's - All MangaStream sources will have to be migrated!

Before, the full url of a series or chapter was set as the id. This is bad for many reasons, the main on being if the source changes it's url or path names even slightly, the id changes, and you will be forced to migrate all saved library entries for that source.

Now, the id is parsed from the url, no longer the full url, this means the chances of a series breaking and requiring migration is far lower. A site could change its url and pathnames all they want, and no migration would be necessary, all existing library entries will work fine (once the url's and pathnames have been updated).

I also formatted all json files with `jq --tab` so the formatting will finally be consistent

Checklist:


Skittyblock:
since this breaks every mangastream source, I'm thinking I might move the in-app migration tool to the top of my priority list and get that done quick before merging...

Skittyblock:
I think you could also update `aidoku-rs` for mangastream while you're at it? I believe it's using the outdated version, just delete `Cargo.lock` and recompile and you'll have to make a few changes.

Skittyblock (dismissed):

PR #137, opened 2023-01-02
HentaiVN Rust rewrite

beerpiss:
Checklist:


src/rust/vi.hentaivn/src/search.rs
```diff
@@ -0,0 +1,209 @@
+use crate::BASE_URL;
+use aidoku::{
+	helpers::uri::QueryParameters,
+	prelude::format,
+	std::{String, Vec},
+	Filter, FilterType,
+};
+use alloc::string::ToString;
+
+fn genre_map(genre: String) -> String {
```
Skittyblock, line 10: why are you mapping these instead of setting the genre id in filters.json?

Skittyblock (approved):

PR #168, opened 2023-01-16
feat(webtoon)!: complete rewrite of webtoon

KaBankz:
Complete re-write of the webtoon's extension.

> **Warning** **BREAKING CHANGE!**: All webtoon entries will have to be migrated, as the extension id changed, and all series id's. I've tested the migrations multiple times and it all worked so it should be safe.

Why? - The old extension scraped the mobile site, which is slow and does not return much for listings, which made it hard for some people to discover new series. In this re-write the desktop site is used which is much faster and returns a huge listing, so it's much better.

Also, I was able to squash some bugs that existed in the previous extension (covers work 🙂).

P.S multi-language is now supported, not just English.
Newly available languages:
`"zh-hant", "th", "id", "es", "fr", "de"`

Checklist:


src/rust/multi.webtoon/src/parser.rs
```diff
@@ -0,0 +1,426 @@
...
+
+		let author_artist = manga_node.select(".author").text().read();
+		let author_artist = author_artist.split('/').collect::<Vec<&str>>();
+
+		let author = String::from(author_artist[0].trim());
+		let mut artist = String::new();
+
+		if author_artist.len() > 1 {
+			artist = String::from(author_artist[1].trim());
+		}
```
Skittyblock, line 60: I don't think it's worth it to parse any of these -- the app doesn't use or expect them, and the extra scraping of info slows it down a bit. Best to leave the details parsing for the manga details function.

src/rust/multi.webtoon/src/parser.rs
```diff
@@ -0,0 +1,426 @@
...
+	for manga in html
+		.select("#content div.challenge_lst > ul > li > a")
+		.array()
+	{
+		let manga_node = manga.as_node().expect("Failed to get manga node");
+		let id = get_manga_id(manga_node.attr("href").read());
+		let cover = manga_node.select("img").attr("src").read();
+		let title = manga_node.select(".subj").text().read();
+		let categories = [manga_node.select(".genre").text().read()].to_vec();
+		let author = manga_node.select(".author").text().read();
```
Skittyblock, line 109: Same here

Skittyblock (changes requested):

Skittyblock (approved):

PR #178, opened 2023-01-21
Mangaworld

daarkdemon:
add two websites: [mangaworld](https://www.mangaworld.so/) and [mangaworldadult](https://www.mangaworldadult.com)

Checklist:


daarkdemon:
Any update on this ????😅

Skittyblock:
why did you update the readme? I don't think it should be changed like this (I'm not a fan of the badges), but even if it was to be changed, it should be done through a separate pr

src/rust/mangaworld/template/src/template.rs
```diff
@@ -0,0 +1,243 @@
...
+		MangaContentRating::Safe
+	};
+	let manga_type = html
+		.select("div.info > div.meta-data.row.px-1 > div:nth-child(5) > a")
+		.text()
+		.read();
+	let viewer = match manga_type.as_str() {
+		"Manga" => MangaViewer::Rtl,
+		"Manhwa" => MangaViewer::Scroll,
+		_ => MangaViewer::Ltr,
```
Skittyblock, line 155: why are you falling back to ltr? you should use rtl for manga sources

Skittyblock (changes requested):
and revert readme changes

src/rust/mangaworld/template/src/template.rs
daarkdemon: It has mostly Chinese comics so it is left to right

Skittyblock:
you still have readme changes

src/rust/mangaworld/template/src/template.rs
Skittyblock: Manhua should be marked as scroll. I looked at the site, and the leftover types are like Doujinshi and Oneshot, which should also be Rtl.

Skittyblock (approved):

bordam:
@daarkdemon Thank you so much!!! 😊 Can you share a link to your PayPal? Or can I donate to you in other ways?

PR #190, opened 2023-01-26
Update NHentai extension

choboro:
Checklist:

Search was broken sometimes so I tried to fix it by changing how the language list is handled. Seems to only be fixable with newer ArrayRef. Also added Tomboys into the filter list for justice.


src/rust/multi.nhentai/src/lib.rs
```diff
@@ -87,7 +81,7 @@ fn get_manga_list(filters: Vec<Filter>, page: i32) -> Result<MangaPageResult> {
 	if is_sauce_code {
 		let url = helper::get_details_url(sauce_code);
 		let request = Request::new(&url, HttpMethod::Get).header("User-Agent", "Aidoku");
```
Skittyblock, line 83:
```suggestion
		let request = Request::new(url, HttpMethod::Get).header("User-Agent", "Aidoku");
```

PR #82, opened 2022-09-13
Add source zh.manhuaren

geniusgordon:
Checklist:
- [x] Updated source's version for individual source changes
- [x] Updated all sources' versions for template changes
- [x] Set appropriate `nsfw` value
- [x] Did not change `id` even if a source's name or language were changed
- [x] Tested the modifications by running it on the simulator or a test device

PR #212, opened 2023-02-14
new source: zh.manhuaren

Trzeth:
Note:
- clean up #82 

Checklist:

For test:
[manhuaren.aix](https://github.com/Skittyblock/aidoku-community-sources/files/10728562/manhuaren.zip)


src/rust/zh.manhuaren/src/helper.rs
```diff
@@ -0,0 +1,151 @@
...
+			result.push(b'%');
+			result.push(hex[curr as usize >> 4]);
+			result.push(hex[curr as usize & 15]);
+		}
+	}
+
+	String::from_utf8(result).unwrap_or_default()
+}
+
+pub fn encode_uri(string: String) -> String {
```
Skittyblock, line 33: is there a reason you can't use `encode_uri` from aidoku-rs helpers crate?
```toml
[dependencies]
aidoku = { git = "https://github.com/Aidoku/aidoku-rs", features = ["helpers"] }
```
then
```rs
use aidoku::helpers::uri::encode_uri_component;
```

src/rust/zh.manhuaren/src/helper.rs
```diff
@@ -0,0 +1,151 @@
...
+			result.push(b'%');
+			result.push(hex[curr as usize >> 4]);
+			result.push(hex[curr as usize & 15]);
+		}
+	}
+
+	String::from_utf8(result).unwrap_or_default()
+}
+
+pub fn i32_to_string(mut integer: i32) -> String {
```
Skittyblock, line 63: use alloc ToString instead

src/rust/zh.manhuaren/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+	let url = String::from(API_URL) + "/v1/manga/getDetail?" + &qs;
+
+	let body = helper::request(url, HttpMethod::Get)?;
+
+	let json = json::parse(body)?.as_object()?;
+	let manga = json.get("response").as_object()?;
+	let category_str = manga
+		.get("mangaTheme")
+		.as_string()
+		.unwrap_or_else(|_| StringRef::from(""))
```
Skittyblock, line 98: it's not very good to do this. something better would be
```rs
let category_str = match manga.get("mangaTheme").as_string() {
	Ok(str) => str.read(),
	Err(_) => String::new(),
};

File: src/rust/zh.manhuaren/src/lib.rs [hunk 4]
```diff
@@ -0,0 +1,460 @@
...
+
+		manga_arr.push(Manga {
+			id: helper::i32_to_string(manga_obj.get("mangaId").as_int().unwrap_or(0) as i32),
+			cover:
+				// this cover val won't be empty
+				manga_obj
+				.get("mangaCoverimageUrl")
+				.as_string()
+				.unwrap_or_else(|_| StringRef::from(""))
+				.read(),
```
Skittyblock, line 324: if it will never be empty, you could just `.expect("error message if it doesn't exist")` it

src/rust/zh.manhuaren/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+			artist: manga_obj
+				.get("mangaAuthor")
+				.as_string()
+				.unwrap_or_else(|_| StringRef::from(""))
+				.read(),
+			description: String::new(),
+			url: String::new(),
+			categories: Vec::new(),
```
Skittyblock, line 342: instead of doing this, use `..Default::default()` at the end of the struct to fill in unused properties with default values

src/rust/zh.manhuaren/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+	})
+}
+
+fn parse_chapters(manga: &ObjectRef, key: &str) -> Vec<Chapter> {
+	match manga.get(key).as_array() {
+		Ok(chapters) => {
+			let mut chapter_arr: Vec<Chapter> = Vec::new();
+
+			for ch in chapters {
+				let ch_obj = ch.as_object().unwrap();
```
Skittyblock, line 365:
```suggestion
				let ch_obj = ch.as_object().expect("chapter item should be object");
```
or equivalent error message

src/rust/zh.manhuaren/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+
+				let mut url = helper::encode_uri(String::from(&host));
+				url.push_str(&p_str);
+				url.push_str(&query);
+
+				page_arr.push(Page {
+					index: (i + 1) as i32,
+					url,
+					base64: String::new(),
+					text: String::new(),
```
Skittyblock, line 447: `..Default::default()`

Skittyblock (changes requested):
thanks for reviving a dead pr, here are some more changes
- use tabs for json formatting
- try not to create your own StringRefs just to read them into a rust string (inefficient)
- don't use `i32_to_string`
- don't unwrap (use expect instead if you're sure it'll never fail unless site changes)
- use `Default::default()` for struct default values

Skittyblock:
Yeah checking if it's empty is fine, but you don't need to match, you just can just do an if
```rs
if str.is_empty() {
	fallback
} else {
	str
}
```

The default function is fine, yeah, but you should probably name them something different since `unwrap_or_default` is used when something has the default trait implemented, and the `unwrap_or_empty` function unwraps or returns the fallback, not an empty string, so the name is a little misleading.

Trzeth:
Done. 😊

Skittyblock:
json still using spaces for indentation, make sure it's tabs

Trzeth:
something wrong with vscode builtin formatter, fixed now.

Skittyblock (approved):

PR #273, opened 2023-04-01
feat: rawkuma source

swaggy-p-jp:
Adding in rawkuma - japanese source  and apparently now has some Korean manhwa - to mangastream template. Rawkuma looks to not append the protocol to the images in the html on first load after a GET request so added in a feature to the mangastream source to allow for adding in protocol rather than rewrite the entire page_requests section in the template.

Quick video of the previous source and new source working 

ezgif com-optimize

Checklist:


src/rust/mangastream/template/src/template.rs
```diff
@@ -400,7 +402,7 @@ impl MangaStreamSource {
 		} else {
 			for (at, page) in html.select(self.page_selector).array().enumerate() {
 				let page_node = page.as_node().expect("Failed to get page as node");
-				let page_url = urlencode(page_node.attr(self.page_url).read());
+				let page_url = if self.protocol { format!("{}{}", "https:", urlencode(page_node.attr(self.page_url).read())) } else { urlencode(page_node.attr(self.page_url).read()) } ;
```
Skittyblock, line 405: this shouldn't be all on one line, run `cargo fmt` and it'll format it for you correctly

swaggy-p-jp: Done. Do you think it's worth it to have it run as part of a pre-commit linter?

Skittyblock (approved):

PR #275, opened 2023-04-01
Add,remove, & update various sources

naufaljct48:
Remove sources
- ManhwaIndo (ID)
- Manhwaland (18+) (ID)
They're no longer updating mangas/manhwas

Add new sources:
- Manhwalist (ID)

Update sources:
- KomikTap (18+) (ID)

Checklist:


Skittyblock:
do the sources still work? as long as the websites exist and sources work, I don't think they should be removed if they're just not adding anything new to the site

naufaljct48:
> do the sources still work? as long as the websites exist and sources work, I don't think they should be removed if they're just not adding anything new to the site

The site are still alive, but there's an error with the sources. "Failed to load chapter", even though there's no changes with the sites / sources and i cannot fix that

Skittyblock:
alright if the sources are broken that's fine then

Skittyblock (approved):

PR #282, opened 2023-04-05
fix(asurascans): remove has_permanent_chapter_url

KaBankz:
Unfortunately not all chapters on asura support permanent url's so some chapters break if this is enabled

This also means read history will be lost if asura rerolls the random int prefix (which they have been doing frequently) on the chapter url's as that will change the chapter id

- fix(asurascans): remove has_permanent_chapter_url
- chore(asurascans): bump version

Checklist:

### DIFF
--- a/src/rust/mangastream/sources/asurascans/res/source.json
+++ b/src/rust/mangastream/sources/asurascans/res/source.json
@@ -3,7 +3,7 @@
 		"id": "multi.asurascans",
 		"lang": "multi",
 		"name": "Asura Scans",
-		"version": 5,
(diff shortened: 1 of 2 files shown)


Skittyblock:
is there no way to selectively detect if a chapter supports it? :(

KaBankz:
> is there no way to selectively detect if a chapter supports it? :(

Not that I'm aware of, the chapters that support it and don't look identical in the HTML. The only way to tell, is to actually request the chapter page and see if it returns 404 or 301

Skittyblock (approved):

PR #298, opened 2023-04-11
new source: multi.simplyh

itss0n1c:
Checklist:


src/rust/multi.simplyhentai/res/settings.json
```diff
@@ -0,0 +1,16 @@
...
+  {
+    "type": "group",
+    "title": "Settings",
+    "items": [
+      {
+        "type": "select",
+        "key": "image_quality",
+        "title": "Image Quality",
+        "values": ["full", "small_thumb", "thumb", "giant_thumb"],
+        "titles": ["Full", "Small Thumb", "Thumb", "Giant Thumb"],
```
Skittyblock, line 11: I don't think average users will understand why you're talking about thumbs. You should make this more readable, like Full, Small, Regular, Giant or something

src/rust/multi.simplyhentai/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+use helper::{make_search_url, API_BASE_URL, USER_AGENT};
+use parser::{parse_chapter_list, parse_manga, parse_page_list, parse_search};
+
+#[get_manga_list]
+fn get_manga_list(filters: Vec<Filter>, page: i32) -> Result<MangaPageResult> {
+	let mut manga_title: Option<String> = None;
+	let mut tags: Vec<String> = Vec::new();
+	let mut sort: Option<&str> = None;
+	let mut language = "English";
+	if let Ok(mut languages) = defaults_get("languages")?.as_array() {
```
Skittyblock, line 26: You're only handling one error here, which defeats the purpose. Don't return if `defaults_get` returns an error (don't use `?`).

src/rust/multi.simplyhentai/src/parser.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+pub fn parse_chapter_list(manga_id: String, res: ObjectRef) -> Result<Vec<Chapter>> {
+	let data = res.get("data").as_object()?;
+	let date_updated =
+		data.get("created_at")
+			.as_date("yyyy-MM-dd'T'HH:mm:ss+ss:ss", None, Some("UTC"))?;
+	Ok(Vec::from([Chapter {
+		id: manga_id,
+		title: "Chapter 1".to_string(),
```
Skittyblock, line 103: don't do this, the app does it automatically if you don't have a title

Skittyblock (changes requested):
also use tabs for json indentation

src/rust/multi.simplyhentai/res/settings.json
itss0n1c: sure ok

PR #339, opened 2023-05-19
feat(source): add new source `myreadingmanga`

### LINKED ISSUE #292: MyReadingManga
Source Request
https://myreadingmanga.info/
language: English

aphronyx:
This PR adds a new multiple languages nsfw source [**MyReadingManga**](https://myreadingmanga.info/).

## Checklist

## Related Issues

- Close #292

## Changes

- Added new source `myreadingmanga`

## Notes

This source was tested on both simulator and TestFlight app.


src/rust/multi.myreadingmanga/res/source.json
```diff
@@ -0,0 +1,57 @@
+{
+	"info": {
+		"id": "multi.myreadingmanga",
+		"lang": "multi",
+		"name": "MyReadingManga",
+		"version": 1,
+		"url": "https://myreadingmanga.info/",
+		"nsfw": 2
+	},
+	"_languages": [
```
Skittyblock, line 10: is there a way to add support for filtering by language or no?

aphronyx: yes, but the filter of the site is **AND** not **OR**. if users select multiple languages, it’ll filter out many manga they might actually want. i don’t have android phone, so i’m not sure how tachiyomi does, but [it seems to apply the above filter](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/c6722d44b0e32bb3056f7f9f50884868541b2b1d/src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt#LL75C14-L75C14). 

the search field can exclude keywords users don’t want, but i couldn’t find the full list of languages the site have. plus, some manga have multiple languages in a single manga, using exclude method will also filter them out, e.g. a manga is [Eng/JP], if i don’t select japanese (exclude japanese), this manga will be filtered out. 

i still can’t figure out a better way to filter by language

Skittyblock: ah, it has an enforce language filter. In that case, I would probably add the same filter (which does the strict AND filtering) in filters.json then and ignore the languages thing in source.json.

src/rust/multi.myreadingmanga/res/filters.json
aphronyx, line 1: i found the filter list, but there are too many of them.
should i keep all of them or just some of them?

Skittyblock: just some of them. the app struggles to process a massive amount of filters.

aphronyx: ok! i’ll keep 10 items at most for each filter.

Skittyblock (approved):

PR #414, opened 2023-08-07
feat(boylove): add chinese character sets setting

aphronyx:
This PR adds a new feature that allows users to switch between traditional Chinese and simplified Chinese in `boylove`.

## Checklist

## Changes

- Added Chinese character sets setting
- Improved code readability
- Simplified description processing
- Updated source’s version

## Screenshots

| Setting | Traditional Chinese | Simplified Chinese |
| :-: | :-: | :-: |
| Simulator Screenshot - iPhone 14 Pro - 2023-08-08 at 02 46 20<br>Transl.: Switch to simplified/traditional | Simulator Screenshot - iPhone 14 Pro - 2023-08-08 at 02 48 40 | Simulator Screenshot - iPhone 14 Pro - 2023-08-08 at 02 49 11 |

## Notes


Skittyblock (approved):

src/rust/zh.boylove/res/source.json
```diff
@@ -3,7 +3,7 @@
 		"id": "zh.boylove",
 		"lang": "zh",
 		"name": "香香腐宅（無VIP）",
```
Skittyblock, line 5:
```suggestion
		"name": "香香腐宅",
```

src/rust/zh.boylove/res/settings.json
```diff
@@ -0,0 +1,14 @@
+[
+    {
+        "type": "group",
+        "items": [
+            {
+                "type": "switch",
```
aphronyx, line 6: @Skittyblock 
is `handle_notification` supported for `"type": "segment"`? i tried to use that at first cuz it might be clearer for users, but it did nothing when i tapped the options

```json
{
    "type": "segment",
    "key": "isTC",
    "title": "切換為",
    "notification": "switchChineseCharSet",
    "default": "繁體",
    "values": [
        "简体",
        "繁體"
    ]
}
```

Simulator Screenshot - iPhone 14 Pro - 2023-08-09 at 21 10 58

Skittyblock: should be https://github.com/Aidoku/Aidoku/blob/761e4fdf3feed72d99df21b1781626ef83a46f11/iOS/Old%20UI/Base/Settings/SegmentTableViewCell.swift#L89

aphronyx:
```rust
#[handle_notification]
fn handle_notification(notification: String) {
	println!("notification");

	if notification.as_str() == "switchChineseCharSet" {
		switch_chinese_char_set();
	}
}
```
did i do smth wrong? seemed like the function wasn’t called when i changed the setting

Skittyblock: sorry, just looked into it now and you were right; I had a logic issue so just segment cells wouldn't send notifications. should be fixed in the next update: https://github.com/Aidoku/Aidoku/commit/2870f1875f1fda49c33899d89b18c5cae96364b9

PR #423, opened 2023-08-17
feat(boylove): add sign in setting

aphronyx:
This PR adds a new feature that allows users to sign in to their account in `boylove`.
Users can now sign in to their account and read manga for VIP members if their membership level is VIP.
VIP members can also enable the auto check in setting to keep their VIP membership.

## Checklist

## Changes

- Updated source’s version
- Separated enum `Url` and related functions into module `url.rs`
- Added sign in setting
- Added auto check in setting

## Screenshots

| Get CAPTCHA | Base64-encoded CAPTCHA | Decode Base64 | Sign In |
| :-: | :-: | :-: | :-: |
| Simulator Screenshot - iPhone 14 Pro - 2023-08-19 at 00 30 04 | captcha_base64 | base64_to_image | Simulator Screenshot - iPhone 14 Pro - 2023-08-19 at 00 29 48 |

| Sign In Successfully | Wrong Username or Password | Wrong CAPTCHA |
| :-: | :-: | :-: |
| sign_in_successfully | wrong_username_or_password | wrong_captcha |

| Filter Viewing Permission | Non-VIP | VIP |
| :-: | :-: | :-: |
| filter_viewing_permission | nonvip_search | vip_search |

## Notes


Skittyblock (approved):
not sure I really like the flow of requiring a user to decode a base64 image sent to them via logs, but I don't really see a better way at this point and it makes sense to still have the option to sign in. might add something to the source api to help with this, but idk if it's worth if it's only this one source that would use it.

aphronyx:
> might add something to the source api to help with this, but idk if it's worth if it's only this one source that would use it.

that’s true, i don’t think there’re many sources have different viewing permissions and require captcha when signing in. this is just an optional feature, not an fatal issue, it’s not worth adding a new api only for this source.

> not sure I really like the flow of requiring a user to decode a base64 image sent to them via logs, but I don't really see a better way at this point and it makes sense to still have the option to sign in.

i think that’s a bad flow too😂
i tried to do this at first, but unfortunately, the cover would be cut
```rust
FilterType::Title => {
    let search_str = filter.value.as_string()?.read();

    if search_str == "aidoku:signin" {
        return Ok(Url::SignInPage);
    }
}
```
```rust
let manga_list_url = Url::from(filters, page)?;

if manga_list_url == Url::SignInPage {
	let sign_in_page = manga_list_url.request(HttpMethod::Get).html()?;

	let captcha_img_path = sign_in_page.select("img#verifyImg").attr("abs:src").read();

	let manga = Vec::from([Manga {
		id: "aidoku:signin".to_string(),
		cover: captcha_img_path,
		title: "Sign In".to_string(),
		..Default::default()
	}]);

	let has_more = false;

	return Ok(MangaPageResult { manga, has_more });
}
```
simulator_screenshot_7D0B4A9D-FD58-4773-9545-8E78C1B66C5F

but Aidoku/Aidoku#266 might be a solution to this?

PR #427, opened 2023-08-23
feat(mangastream): postid support

### LINKED ISSUE #382: Asura Scans changing identifiers
I’ve been coming across a bug where whenever a manhwa from Asura Scans updates, it gets added to the Library as a duplicate with reset read/unread information. Below are examples of a particular case:
IMG_2073
IMG_2074
IMG_2075
IMG_2076
I do not know if this issue is specific to Asura Scans or applies to other ones as well, but whenever this happens, as seen in the screenshots, the older version of the source just stays there without the addition of the new chapter.
I’m on version 0.6.3 build 1, iOS 16.5, iPhone 13

KaBankz:
Closes #382

This should solve the issues mangastream sources face whenever the urls or identifiers change. Theoretically, with postids, migration should no longer be necessary now, and chapter history won't be randomly reset.

I tested this with most english mangastream sources, and all was working well, with only 1 or 2 sources that did not support my postids implementation. That's why I have it as an opt-in setting for now, in the future this could become the default behavior if the majority of mangastream sources support it.

yes, it is a bit janky, but it works

Checklist:

### DIFF
--- a/src/rust/mangastream/template/src/helper.rs
+++ b/src/rust/mangastream/template/src/helper.rs
@@ -1,12 +1,16 @@
 use aidoku::{
+	helpers::substring::Substring,
 	prelude::format,
-	std::defaults::defaults_get,
-	std::html::Node,
+	std::{current_date, html::Node},
+	std::{defaults::defaults_get, net::Request},
 	std::{String, StringRef, Vec},
 	MangaStatus,
 };
 
 use crate::template::MangaStreamSource;
+
+extern crate hashbrown;
+use hashbrown::HashMap;
 
 // generate url for listing page
 pub fn get_listing_url(
@@ -339,6 +343,16 @@ pub fn get_id_from_url(url: String) -> String {
 		url.pop();
 	};
 
+	// if there is a post id in the url, return it
+	if url.contains("p=") {
+		return String::from(
+			url.substring_after("p=")
+				.expect("Failed to parse id from url")
+				.substring_before("&")
+				.expect("Failed to parse id from url"),
+		);
+	}
+
 	// this will get the last part of the url
 	// example https://flamescans.org/series/the-world-after-the-fall
 	// will return the-world-after-the-fall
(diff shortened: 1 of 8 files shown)


src/rust/mangastream/template/src/helper.rs
```diff
@@ -359,3 +373,105 @@ pub fn get_lang_code() -> String {
...
+				return mapping.clone();
+			}
+		}
+	}
+
+	let all_manga_listing_url = format!("{}/{}/list-mode", url, pathname);
+
+	let html = Request::get(all_manga_listing_url)
+		.html()
+		.expect("Failed to get html");
```
Skittyblock, line 403: probably best to return a result for this function and fail gracefully since this could fail and it would break the extension until the app is relaunched.

src/rust/mangastream/template/src/helper.rs
```diff
@@ -359,3 +373,105 @@ pub fn get_lang_code() -> String {
...
+}
+
+/// Search the `MANGA_URL_TO_POSTID_MAPPING` for the postid from a manga url
+pub fn get_postid_from_manga_url(url: String, base_url: &str, pathname: &str) -> String {
+	let manga_url_to_postid_mapping = generate_manga_url_to_postid_mapping(base_url, pathname);
+
+	String::from(
+		manga_url_to_postid_mapping
+			.get(&url)
+			.expect("Failed to get post id from url"),
```
Skittyblock, line 431: same here

src/rust/mangastream/template/src/template.rs
```diff
@@ -232,7 +248,13 @@ impl MangaStreamSource {
...
 	// parse manga details page
 	pub fn parse_manga_details(&self, id: String) -> Result<Manga> {
-		let url = format!("{}/{}/{}", self.base_url, self.traverse_pathname, id);
+		let url = {
+			if self.use_postids {
+				format!("{}/{}/?p={}", self.base_url, self.traverse_pathname, id)
+			} else {
+				format!("{}/{}/{}", self.base_url, self.traverse_pathname, id)
+			}
+		};
```
Skittyblock, line 257: don't need to wrap this in an extra block

src/rust/mangastream/template/src/template.rs
```diff
@@ -331,18 +369,36 @@ impl MangaStreamSource {
...
-					get_permanet_url(original_url)
-				} else {
-					original_url
-				}
-			};
-			let chapter_id = get_id_from_url(chapter_url.clone());
+				chapter_id = String::from(
+					chapter_url_to_postid_mapping
+						.get(&original_url)
+						.expect("Failed to get chapter id from url mapping"),
```
Skittyblock, line 382: probably just `?` here

src/rust/mangastream/template/src/template.rs
```diff
@@ -356,17 +412,22 @@ impl MangaStreamSource {
...
 
-	//parse the maga chapter images list
+	//parse the manga chapter images list
 	pub fn parse_page_list(&self, id: String) -> Result<Vec<Page>> {
 		let url = {
-			let mut url = format!("{}/{}", self.base_url, id);
+			if self.use_postids {
+				format!("{}/?p={}", self.base_url, id)
+			} else {
+				let mut url = format!("{}/{}", self.base_url, id);
```
Skittyblock, line 421: this doesn't have to be a mutable variable, just do an if else

src/rust/mangastream/template/src/helper.rs
```diff
@@ -359,3 +373,105 @@ pub fn get_lang_code() -> String {
...
+		.header("Referer", base_url)
+		.html()
+		.expect("Failed to get html");
+
+	// Janky retry logic to bypass rate limiting
+	// Retry after 10 seconds if we get rate limited. 10 seconds is the shortest
+	// interval that we can retry without getting rate limited again.
+	if html.select("title").text().read() == "429 Too Many Requests" {
+		loop {
+			if start + 10.0 < current_date() {
```
Skittyblock, line 459: is there a way to sleep or something instead of spamming the imported function? there's overhead transferring the result from the app to the source

Skittyblock (changes requested):

src/rust/mangastream/template/src/helper.rs
KaBankz: Not that I'm aware of

src/rust/mangastream/template/src/helper.rs
KaBankz: Had to have an `AidokuError` and none of them seemed appropriate, so I just went with `AidokuErrorKind::Unimplemented`

src/rust/mangastream/template/src/template.rs
KaBankz: Had to have an `AidokuError` and none of them seemed appropriate, so I just went with `AidokuErrorKind::Unimplemented`

Skittyblock (approved):
lgtm

Issue #332, opened 2023-05-12
mangairo


baraa272:
https://w.mangairo.com/home

1.doesnt have cloudflare so easier 
2.doesnt change manga Id like how mangakakalot and mangabat do 
3. Very stable site 
4.fast on updates 
And thank you so much for this app dev
With all love dev

baraa272:
Thanks very much

PR #447, opened 2023-09-05
Add New Source: MangaIro

getBoolean:
Checklist:

Note: Filters do not work with search
Homepage: https://w.mangairo.com/home

Closes https://github.com/Skittyblock/aidoku-community-sources/issues/332


getBoolean:
I noticed pagination isn't working, I will work on fixing that.

src/rust/en.mangairo/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+	let total_results = parser::parse_manga_list(html, &mut result);
+
+	if let Some(total_results_value) = total_results {
+		if total_results_value > result.len() as i32 * page {
+			return Ok(MangaPageResult {
+				manga: result,
+				has_more: true,
+			});
+		}
+	}
```
Skittyblock, line 32: instead of doing all this, I think you could do something like
```rust
let has_more = total_results.map_or(false, |value| value > result.len() as i32 * page);
```

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,376 @@
...
+	total_str = total_str.replace("Total: ", "");
+	total_str = total_str.replace(" stories", "");
+	total_str = total_str.chars().filter(|&c| c != ',').collect();
+
+	let total_result = total_str.parse::<i32>();
+	if let Ok(total) = total_result {
+		Some(total)
+	} else {
+		None
+	}
```
Skittyblock, line 40:
```suggestion
	total_str.parse::<i32>().ok()
```

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,376 @@
...
+}
+
+pub fn get_chapter_list(html: Node) -> Result<Vec<Chapter>> {
+	let mut chapters: Vec<Chapter> = Vec::new();
+	for chapter in html.select(".chapter_list ul li a").array() {
+		let obj = chapter.as_node().expect("node array");
+		let url = obj.attr("href").read();
+		let id = parse_incoming_url_chapter_id(url.clone());
+
+		if let Some(id_value) = id {
```
Skittyblock, line 128: You don't need to rename things when you unwrap them
```suggestion
		if let Some(id) = id {
```

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,376 @@
...
+			let split = id_value.split('-');
+			let vec = split.collect::<Vec<&str>>();
+			let chap_num = vec[vec.len() - 1].parse().unwrap();
```
Skittyblock, line 131: just call this `chapter`, and instead of unwrapping you should probably just use `?` to return or fallback to `-1` for no chapter number

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,376 @@
...
+			..Default::default()
+		});
+	}
+	Ok(pages)
+}
+
+pub fn get_filtered_url(filters: Vec<Filter>, page: i32, url: &mut String) {
+	let mut is_searching = false;
+	let mut search_string = String::new();
+	url.push_str("https://w.mangairo.com");
```
Skittyblock, line 164: probably use `BASE_URL` here instead?

Skittyblock (changes requested):
- icon looks pretty low quality, probably find a better one or use https://github.com/tachiyomiorg/tachiyomi-extensions/blob/651e5cf50754700419c190368a81762e71fcb077/multisrc/overrides/mangabox/mangairo/res/web_hi_res_512.png
- jsons should have newlines at the end of the file
- run `cargo clippy` and fix the warnings and then `cargo fmt` to format

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,376 @@
...
+		if pos == 1 {
+			digit *= -1;
+		}
+		string.insert(pos, char::from_u32((digit as u32) + ('0' as u32)).unwrap());
+		integer /= 10;
+	}
+	string
+}
+
+pub fn urlencode(string: String) -> String {
```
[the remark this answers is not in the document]
getBoolean: This URL encode is copied from the website's js code and converted to rust to mimic the same functionality. I do not believe it supports traditional url encoding.

src/rust/en.mangairo/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+
+mod parser;
+use parser::{BASE_URL, USER_AGENT};
+
+#[get_manga_list]
+fn get_manga_list(filters: Vec<Filter>, page: i32) -> Result<MangaPageResult> {
+	let mut result: Vec<Manga> = Vec::new();
+
+	let mut url = String::new();
+	parser::get_filtered_url(filters, page, &mut url);
```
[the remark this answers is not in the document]
getBoolean: No particular reason, much of lib.rs was copied from the mangapill extension.

I like that much better, I will do that.

src/rust/en.mangairo/src/lib.rs
getBoolean: Thanks! I knew there had to be a better way but wasn't sure how to do it in Rust

src/rust/en.mangairo/src/parser.rs
getBoolean: I've updated it to use `unwrap_or(-1.0)`

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,376 @@
...
+		.find(|&x| x.kind == FilterType::Select && x.name == "Status")
+		.cloned();
+	let sort_filter: Option<Filter> = filters
+		.iter()
+		.find(|&x| x.kind == FilterType::Select && x.name == "Sort")
+		.cloned();
+	let genre_filter: Option<Filter> = filters
+		.iter()
+		.find(|&x| x.kind == FilterType::Select && x.name == "Genre")
+		.cloned();
```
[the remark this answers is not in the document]
getBoolean: I initially did this so I could ensure the URL was constructed in the correct order. I've refactored this function to use match for the filters and pull out the info needed. I've also removed the mutable strings in this function except for `search_string` and refactored the filter matches to support that.

```rust
	let mut title_filter = String::new();
	let mut author_filter = String::new();
	let mut sort_filter = -1;
	let mut genre_filter = -1;
	let mut status_filter = -1;
	for filter in filters {
		match filter.kind {
			FilterType::Title => {
				if let Ok(filter) = filter.value.as_string() {
					title_filter = urlencode(filter.read().to_lowercase());
					is_searching = true;
				}
			}
			...
```

src/rust/en.mangairo/src/parser.rs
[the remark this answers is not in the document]
getBoolean: For reference, URI escaping the space yields no results but it does work with underscores.

* https://w.mangairo.com/list/search/solo%20leveling
* https://w.mangairo.com/list/search/solo_leveling

getBoolean:
Thanks for the thorough review! I especially appreciate the small tips you gave that might otherwise have been overlooked by others.

I've addressed all the feedback comments, but I'd like you to take a look at the fixes to make sure I addressed them properly.

getBoolean:
I scaled down that icon you linked to 128x128 from 512x512. The app currently only supports 128x128, correct?

src/rust/en.mangairo/src/parser.rs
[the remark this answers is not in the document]
getBoolean: I am not sure if the mutable variables approach is the best way to do this, but I'm not aware of a better option

src/rust/en.mangairo/res/settings.json
Skittyblock, line 1: no need to have this empty file, you can just remove it

src/rust/en.mangairo/res/source.json
```diff
@@ -0,0 +1,25 @@
+{
+	"info": {
+		"id": "en.mangairo",
+		"lang": "en",
+		"name": "MangaIro",
+		"version": 1,
+		"url": "https://w.mangairo.com/home",
```
Skittyblock, line 7: In order for deep links to work and such this should be the base url, even if it isn't the home listing page
```suggestion
		"url": "https://w.mangairo.com",
```

src/rust/en.mangairo/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+		.header("Referer", BASE_URL)
+		.header("User-Agent", USER_AGENT);
+}
+
+#[handle_url]
+fn handle_url(url: String) -> Result<DeepLink> {
+	let parsed_manga_id = parser::parse_incoming_url_manga_id(url.clone());
+	let parsed_chapter_id = parser::parse_incoming_url_chapter_id(url);
+
+	if parsed_manga_id.is_none() {
```
Skittyblock, line 72: typically best to unwrap by doing something like, rather than calling `unwrap()`
```rust
if let Some(parsed_manga_id) = parsed_manga_id {
    Ok(DeepLink { ... })
} else {
    Err(...)
}
```

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,354 @@
...
+				..Default::default()
+			});
+		}
+	}
+
+	// Example: 'Total: 38,202 stories'
+	let mut total_str = html.select(".quantitychapter").text().read();
+	total_str = total_str.replace("Total: ", "");
+	total_str = total_str.replace(" stories", "");
+	total_str = total_str.chars().filter(|&c| c != ',').collect();
```
Skittyblock, line 35: this doesn't need to be mutable, you could do all the calls on one variable (including the i32 parse)

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,354 @@
...
+		.read()
+		.trim()
+		.to_string();
+	let status_str = html
+		.select(".story_info_right li:nth-child(5) a")
+		.text()
+		.read()
+		.to_lowercase();
+
+	let url = format!("{}", &id);
```
Skittyblock, line 62: ids should **not** be full urls. the best way to handle this is to string the url out so you just have the `/relative/path`, and then you can prepend the `BASE_URL` to get the full url. This makes it easier to fix if the site ever changes domains.

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,354 @@
...
+	let mut chapters: Vec<Chapter> = Vec::new();
+	for chapter in html.select(".chapter_list ul li a").array() {
+		let obj = chapter.as_node().expect("node array");
+		let url = obj.attr("href").read();
+		let id = parse_incoming_url_chapter_id(url.clone());
+
+		if let Some(id) = id {
+			let split = id.split('-');
+			let vec = split.collect::<Vec<&str>>();
+			let chapter: f32 = vec[vec.len() - 1].parse().unwrap_or(-1.0);
```
Skittyblock, line 131: I think you can do something like this instead
```suggestion
			let chapter = id
				.rsplit_once('-')
				.map(|v| v.1.parse::<f32>().ok())
				.flatten()
				.unwrap_or(-1.0);
```

src/rust/en.mangairo/src/parser.rs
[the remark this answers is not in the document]
Skittyblock: oh okay, fair enough

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,354 @@
...
+	str = str.replace(match_y, "y");
+	str = str.replace(match_d, "d");
+	str = str.replace(match_symbols, "_");
+	str = replace_consecutive_underscores(str);
+	str = str.trim_matches('_').to_string();
+
+	str
+}
+
+fn replace_consecutive_underscores(input: String) -> String {
```
Skittyblock, line 337: I'm a little confused. Isn't this the same as `.replace("__", "_")`?

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,354 @@
...
+
+	let mut authors: Vec<String> = Vec::new();
+	html.select(".story_info_right li:nth-child(3) a")
+		.array()
+		.for_each(|tag| {
+			authors.push(String::from(
+				tag.as_node().expect("node array").text().read().trim(),
+			))
+		});
+	let author = authors.join(", ");
```
Skittyblock, line 72: I think you could do this
```suggestion
	let author = html
		.select(".story_info_right li:nth-child(3) a")
		.array()
		.map(|tag| {
			String::from(
				tag.as_node().expect("node array").text().read().trim(),
			)
		})
		.join(", ");
```

src/rust/en.mangairo/src/parser.rs
```diff
@@ -0,0 +1,354 @@
...
+
+	let mut categories: Vec<String> = Vec::new();
+	html.select(".story_info_right .a-h")
+		.array()
+		.for_each(|tag| categories.push(tag.as_node().expect("node array").text().read()));
```
Skittyblock, line 77: again, just map these instead of pushing to a mutable vector

Skittyblock (changes requested):

Skittyblock:
> I scaled down that icon you linked to 128x128 from 512x512. The app currently only supports 128x128, correct?

the app supports anything, but we use 128x128 conventionally since that's the largest the source icon would ever need to be and it's best to be consistent.

src/rust/en.mangairo/src/parser.rs
getBoolean: That would collapse three underscores `___` to `_`? I was thinking it would not, but now that I think about it again it makes sense it would.

src/rust/en.mangairo/src/parser.rs
getBoolean: I ran into a problem with that because the latest manga use URL `w.mangairo.com`, but the older manga use `chap.mangairo.com`. I don't remember exactly what was the problem I ran into, I'll revisit this again.

src/rust/en.mangairo/res/settings.json
Skittyblock: still needs to be removed

src/rust/en.mangairo/src/parser.rs
Skittyblock: is it possible to only use `chap.mangairo.com` then?

src/rust/en.mangairo/res/source.json
```diff
@@ -0,0 +1,25 @@
+{
+	"info": {
+		"id": "en.mangairo",
+		"lang": "en",
+		"name": "MangaIro",
+		"version": 1,
+		"url": "https://w.mangairo.com",
```
Skittyblock, line 7: In order to get deeplinks to work
```suggestion
		"urls": [
			"https://w.mangairo.com",
			"https://chap.mangairo.com"
		],
```

Skittyblock (changes requested):

src/rust/en.mangairo/src/parser.rs
getBoolean: Should be... I'll need to do some refactoring though.

getBoolean: Unfortunately, it doesn't work. `https://chap.mangairo.com/story-de293390/chapter-1` will redirect to `https://w.mangairo.com/story-de293390`, breaking method `get_page_list`

Skittyblock: alright then it's probably fine to just leave it with the full urls

Skittyblock (approved):

PR #502, opened 2023-11-01
fix(copymanga): chapters sorted incorrectly

aphronyx:
This PR fixes the bug that chapters are sorted incorrectly.

## Checklist

## Premise

Chapters in the same group are already sorted in the JSON file; but the current sorting method (sort by date updated) will mess up the correct order.

## Changes

- Updated dependencies in `Cargo.lock`
- Prevented chapters in the same group from being sorted
- Prefixed group name to chapter title
- Bumped version

## Screenshots

| Before | After |
| :-: | :-: |
| chapters_before | chapters_after |
| version_before | version_after |

## Notes

### DIFF
--- a/src/rust/zh.copymanga/Cargo.lock
+++ b/src/rust/zh.copymanga/Cargo.lock
@@ -100,9 +100,9 @@ dependencies = [
 
 [[package]]
 name = "cpufeatures"
-version = "0.2.9"
+version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a17b76ff3a4162b0b27f354a0c87015ddad39d35f9c0c36607a3bdd175dde1f1"
+checksum = "ce420fe07aecd3e67c5f910618fe65e94158f6dcc0adf44e00d69ce2bdfe0fd0"
 dependencies = [
  "libc",
 ]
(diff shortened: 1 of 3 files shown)


Skittyblock:
How come you just use an index for the chapter numbers rather than actually using the chapter number in the title? The ideal behavior is to strip out the number from the title, pass that number to aidoku as the chapter number, and then just pass a title back if there actually is one other than just the "Chapter X"

aphronyx:
> How come you just use an index for the chapter numbers rather than actually using the chapter number in the title? The ideal behavior is to strip out the number from the title, pass that number to aidoku as the chapter number, and then just pass a title back if there actually is one other than just the "Chapter X"

i’ll rewrite the source and try to do that after the new api is out, but they don’t seem to follow the same naming conventions, making it difficult to parse the number:
| 第 X 回 (Chapter X) | 全一卷 (Only One Volume) | 第 X 話 (Chapter X in Traditional Chinese) | 第 X 话 (Chapter X in Simplified Chinese) |
| :-: | :-: | :-: | :-: |
| Screenshot 2023-11-02 at 23 14 37 | Screenshot 2023-11-02 at 23 15 03 | Screenshot 2023-11-02 at 23 15 17 | Screenshot 2023-11-02 at 23 15 44 |

Skittyblock:
yeah once the new api is out and sources are being rewritten I'll review sources a little closer now that I know everything to look for so I can provide some better feedback before they go through. I'm sure there's a good way to do it here.

Skittyblock (approved):

aphronyx:
to future me,
not the best, but apply to the most common naming conventions:

```rust
use regex::Regex;

pub trait Title {
	fn parse_title(&self) -> (f32, f32, String);
}

impl Title for String {
	fn parse_title(&self) -> (f32, f32, String) {
		if self == "全一卷" {
			return (1.0, -1.0, String::new());
		}

		let re = Regex::new(
			r"^(单行本：)*(第*(?<volume>\d+)[卷部])*(第*(?<chapter>\d+|\d+(\.|-)\d+)[話话回])*(\s|$)(?<title>.*)",
		)
		.unwrap();
		let Some(caps) = re.captures(self.trim()) else {
			return (-1.0, -1.0, self.to_string());
		};

		let volume = caps
			.name("volume")
			.and_then(|m| m.as_str().parse().ok())
			.unwrap_or(-1.0);
		let chapter = caps
			.name("chapter")
			.and_then(|m| m.as_str().replace('-', ".").parse().ok())
			.unwrap_or(-1.0);
		let title = if chapter == -1.0 && volume == -1.0 {
			self.to_string()
		} else {
			caps.name("title")
				.map(|m| m.as_str().to_string())
				.unwrap_or_default()
		};

		(volume, chapter, title)
	}
}
```

PR #514, opened 2023-11-25
Add New source: QuestionableContent.net

Shirk:
This PR adds a source for [Questionable Content.net](https://www.questionablecontent.net) - a web comic running for 20 odd years and still going.

_Note: Currently the chapters do not have an upload-date as those are only ever available for the current year and I'd rather not include any than having to offer a rough guess on a date._

Checklist:
- [x] ~~Updated all sources' versions for template changes~~ (NA)


Shirk:
I admit, I forgot to run clippy in linting mode before pushing my initial version - should be clean now.
buld.sh is also has the executable bit set as well, probably an oversight from using an old version of aidoku-cli at first and locally just running via zsh ./build.sh ^^

Skittyblock (approved):

PR #534, opened 2023-12-16
Update to the new Url for ComicK

### LINKED ISSUE #533: Comick changed domain
They changed from https://comick.app/ to https://comick.cc/

HeapUnderfl0w:
Checklist:

ComicK moved from their `.app` domain to https://comick.cc (and https://api.comick.fun for the API).

Was unable to test the source, as the aidoku-cli just segfaults on my machine.
Should not cause any issues tho, as this is purely a url change in the strings.

Fixes #533

### DIFF
--- a/src/rust/multi.comick/src/lib.rs
+++ b/src/rust/multi.comick/src/lib.rs
@@ -11,7 +11,7 @@ pub mod helper;
 pub mod parser;
 
 const BASE_URL: &str = "https://comick.cc";
-const API_URL: &str = "https://api.comick.fun";
+const API_URL: &str = "https://api.comick.cc";
 
 #[get_manga_list]
 fn get_manga_list(filters: Vec<Filter>, page: i32) -> Result<MangaPageResult> {
--- a/src/rust/multi.comick/src/lib.rs
+++ b/src/rust/multi.comick/src/lib.rs
@@ -10,8 +10,8 @@ use aidoku::{
 pub mod helper;
 pub mod parser;
 
-const BASE_URL: &str = "https://comick.app";
-const API_URL: &str = "https://api.comick.app";
+const BASE_URL: &str = "https://comick.cc";
+const API_URL: &str = "https://api.comick.fun";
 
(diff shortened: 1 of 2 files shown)


src/rust/multi.comick/src/lib.rs
```diff
@@ -10,8 +10,8 @@ use aidoku::{
 pub mod helper;
 pub mod parser;
 
-const BASE_URL: &str = "https://comick.app";
-const API_URL: &str = "https://api.comick.app";
+const BASE_URL: &str = "https://comick.cc";
+const API_URL: &str = "https://api.comick.fun";
```
Skittyblock, line 14:
```suggestion
const API_URL: &str = "https://api.comick.cc";
```

Skittyblock (approved):

src/rust/multi.comick/src/lib.rs
HeapUnderfl0w: The API Docs specifically mention to use `https://api.comick.fun`.

https://api.comick.fun/docs/static/index.html
image

Should this still be .cc or do we respect the docs?

Skittyblock:
> aidoku-cli just segfaults on my machine

you need to download aidoku-cli from the repo. I'm guessing you're using the version on brew, which doesn't work.

also that doesn't prevent you from testing; you can just compile the .aix and then install it on your device.

HeapUnderfl0w:
Ah right, didnt know that you could install .aix standalone, will do that next time.

HeapUnderfl0w:
The API docs of ComicK mention specifically to use `https://api.comick.fun` as the api root.

The .cc will work fine i guess, just feel like i should at least mention it:
image
https://api.comick.fun/docs/static/index.html

Skittyblock:
I think it means you have to use `api.comick.cc` and not `comick.cc` -- likely a typo/outdated. I'm guessing that the .fun domain will likely stop working at some point since they've migrated from it, so the .cc domain is a better choice for now.

PR #536, opened 2023-12-24
fix(nepnep): unescape html entities in manga title

GameFuzzy:
Checklist:


src/rust/en.nepnep/res/source.json
Skittyblock, line 1: formatting should be reverted, just bump the version

Skittyblock (approved):

PR #541, opened 2024-01-05
update neoxscans(now nexoscans)

julioreigen:
They changed their website name and URL, with this PR the source will work normally.
New URL: http://nexoscans.com

Checklist:


src/rust/madara/sources/nexoscans/res/source.json
```diff
@@ -1,10 +1,10 @@
 {
 	"info": {
-		"id": "pt-br.neoxscans",
+		"id": "pt-br.nexoscans",
```
Skittyblock, line 3: > Did not change id even if a source's name or language were changed

even thought the name changed, the id should stay the same that people can still update their old source.

julioreigen: Done!

Skittyblock (approved):

PR #544, opened 2024-01-08
removed the random aidoku cli exe that i had

AGoodNub:
Checklist:


Skittyblock:
It doesn't look like the exe file actually got removed in this PR. I think you need to reset your branch and then make your changes so that there's just one commit, rather than reusing your branch from your old PR.

Skittyblock:
fixed in 0a6a7ee38582264a51eb1bdee42bec9daba0b73f

PR #562, opened 2024-01-22
fix(comick): genres not loading

### LINKED ISSUE #530: ComicK missing manga metadata
https://github.com/Skittyblock/aidoku-community-sources/blob/3a95a55bcd00c2f1c3e860eafc33dbfea027f6ff/src/rust/multi.comick/src/parser.rs#L221C6-L221C6

When parsing the listing, only the title is loaded, this ends up with all of the comick manga being empty. No further information available even after refreshing metadata.

daarkdemon:
Checklist:

Fixes #530

### DIFF
--- a/src/rust/multi.comick/src/lib.rs
+++ b/src/rust/multi.comick/src/lib.rs
@@ -10,7 +10,7 @@ use aidoku::{
 pub mod helper;
 pub mod parser;
 
-const BASE_URL: &str = "https://comick.cc";
+const BASE_URL: &str = "https://comick.io";
 const API_URL: &str = "https://api.comick.fun";
 
 #[get_manga_list]
--- a/src/rust/multi.comick/src/lib.rs
+++ b/src/rust/multi.comick/src/lib.rs
@@ -10,8 +10,8 @@ use aidoku::{
 pub mod helper;
 pub mod parser;
 
-const BASE_URL: &str = "https://comick.cc";
-const API_URL: &str = "https://api.comick.cc";
+const BASE_URL: &str = "https://comick.fun";
+const API_URL: &str = "https://api.comick.fun";
 
 #[get_manga_list]
 fn get_manga_list(filters: Vec<Filter>, page: i32) -> Result<MangaPageResult> {
--- a/src/rust/multi.comick/src/lib.rs
+++ b/src/rust/multi.comick/src/lib.rs
@@ -10,7 +10,7 @@ use aidoku::{
 pub mod helper;
 pub mod parser;
 
-const BASE_URL: &str = "https://comick.fun";
+const BASE_URL: &str = "https://comick.cc";
 const API_URL: &str = "https://api.comick.fun";
 
 #[get_manga_list]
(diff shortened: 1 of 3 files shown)


src/rust/multi.comick/src/parser.rs
```diff
@@ -148,7 +148,7 @@ pub fn parse_manga_list(
 		}
 		Ok(MangaPageResult {
 			manga: mangas,
-			has_more: false,
+			has_more: true,
```
Skittyblock, line 151: this shouldn't always be true, otherwise it'll loop infinitely. you need to have a check (e.g. if `mangas` is empty) to see if there are more or not.

Skittyblock (changes requested):
in the future I would probably enforce not using `expect`, since it should only ever be used if you are fully certain it should never be triggered, and since sites change so much it can happen, but it's fine for now until 0.7.

src/rust/multi.comick/src/parser.rs
daarkdemon: fiixed please check now

src/rust/multi.comick/src/lib.rs
```diff
@@ -10,8 +10,8 @@ use aidoku::{
 pub mod helper;
 pub mod parser;
 
-const BASE_URL: &str = "https://comick.cc";
-const API_URL: &str = "https://api.comick.cc";
+const BASE_URL: &str = "https://comick.fun";
+const API_URL: &str = "https://api.comick.fun";
```
Skittyblock, line 14: why did you change back to the old `.fun` domain? `.cc` is the new one

src/rust/multi.comick/src/parser.rs
```diff
@@ -106,12 +106,12 @@ pub fn parse_manga_list(
...
 		);
 		let mut mangas: Vec<Manga> = Vec::new();
-		let json = Request::new(&url, HttpMethod::Get)
+		let json = Request::new(url, HttpMethod::Get)
 			.json()
 			.expect("Failed to load JSON")
 			.as_array()
 			.expect("Failed to get json as array");
-		for data in json {
+		for data in json.clone() {
```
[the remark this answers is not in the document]
daarkdemon: Still the clone() will be used for that variable ?

src/rust/multi.comick/src/lib.rs
daarkdemon: It is mentioned in API docs to use  `.fun` for api and `.cc` for website

image

src/rust/multi.comick/src/parser.rs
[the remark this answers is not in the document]
daarkdemon: Resolved

src/rust/multi.comick/src/lib.rs
Skittyblock: shouldn't `BASE_URL` be `.cc`?

daarkdemon:
@Skittyblock any update?

src/rust/multi.comick/src/lib.rs
```diff
@@ -11,7 +11,7 @@ pub mod helper;
 pub mod parser;
 
 const BASE_URL: &str = "https://comick.cc";
```
Skittyblock, line 13: #598 
```suggestion
const BASE_URL: &str = "https://comick.io";
```

Skittyblock (approved):

PR #581, opened 2024-02-18
fix(copymanga): search not working

aphronyx:
This PR fixes the bug that search is not working.

## Checklist

## Changes

- Added extension `.toml` to `config`
- Updated dependencies
- Updated searching url
- Extracted chapter and volume from title
- Bumped version

## Screenshots

| | Before | After |
| :-: | :-: | :-: |
| **Version** | version(before) | version(after) |
| **Search** | search(before) | search(after) |
| **Chapters** | chapters(before) | chapters(after) |

## Notes


Skittyblock (approved):
might be better to use `filter_map` in the future

PR #582, opened 2024-02-19
feat(boylove): get correct volume and chapter

aphronyx:
This PR corrects the volume number and chapter number of a chapter.

## Checklist

## Changes

- Added extension `.toml` to `config`
- Used hard tabs in `settings.json`
- Updated dependencies
- Updated filters
- Updated URLs for deep linking
- Extracted chapter and volume from title
- Bumped version

## Screenshots

| | Before | After |
| :-: | :-: | :-: |
| **Filters** | filters(before) | filters(after) |
| **Chapters** | chapters(before) | chapters(after) |
| **Version** | version(before) | version(after) |

## Notes


src/rust/zh.boylove/res/source.json
Skittyblock, line 1: why did you add trailing spaces to the urls?

also, I didn't realize before, but you shouldn't have the `url` parameter and `urls` parameter at the same time, since it'll just use `url` by default. It automatically uses the first url from `urls` as the base url for opening the web view.

aphronyx: > why did you add trailing spaces to the urls?

i forgot to remove them when pasting

> you shouldn't have the `url` parameter and `urls` parameter at the same time

i thought `url` was essential. i'll correct other sources later

PR #590, opened 2024-02-26
fix:(Suryatoon): Chapter pages not loading + Icon

Kareylo:
Checklist:


Kareylo:
Fixed the `shared reference of mutable static is discouraged` on the template, should be nice for next releases of the template

Skittyblock (changes requested):
you didn't change anything that would really affect the rest of the sources (just resolved a warning), so it doesn't make sense to bump the versions of all of them I don't think

Kareylo:
> you didn't change anything that would really affect the rest of the sources (just resolved a warning), so it doesn't make sense to bump the versions of all of them I don't think

Sounds logic to me.

Skittyblock (approved):

PR #595, opened 2024-02-29
new source: omegascans

Aeonss:
Checklist:


src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,406 @@
...
+			}
+			_ => continue,
+		}
+	}
+
+	if !genres.is_empty() {
+		genres.pop();
+	}
+
+	let url = format!("https://api.omegascans.org/query?query_string={}&series_status={}&order=desc&orderBy=total_views&series_type=Comic&page=1&perPage=1000&tags_ids=[{}]", search_query, status, genres);
```
Skittyblock, line 51: ideally use a `BASE_API_URL` constant like the `BASE_URL` constant since you use it multiple times

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,406 @@
...
+
+	let json = Request::new(url, HttpMethod::Get)
+		.json()
+		.expect("Failed to load JSON")
```
Skittyblock, line 55: don't use `expect` or `unwrap` unless you know for a fact it won't ever be triggered. use `?` to return the error instead.

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,406 @@
...
+
+		let url = format!("{}/series/{}", base_url, slug);
+
+		mangas.push(Manga {
+			id: slug,
+			cover: cover,
+			title: title,
+			author: String::new(),
+			artist: String::new(),
+			description: String::new(),
```
Skittyblock, line 113: use `..Default::default()` at the end of the struct to populate default values instead of including your own

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,406 @@
...
+						1 => {
+							genres.push_str(filter_id.read().as_str());
+							genres.push(',');
+						}
+						_ => continue,
+					}
+				}
+			}
+			FilterType::Select => match filter.name.as_str() {
+				"Status" => match filter.value.as_int().unwrap_or(-1) {
```
Skittyblock, line 32: isn't this only single select? it would make more sense to use `QueryParameters`, or at the very least assign the string instead of pushing to it.

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,406 @@
...
+			.expect("Failed to get manga slug as str")
+			.read();
+
+		let status = manga
+			.get("status")
+			.as_string()
+			.expect("Failed to get manga status as array")
+			.read();
+
+		let mut manga_status = MangaStatus::Unknown;
```
Skittyblock, line 93: this doesn't need to be mutable, just assign it to the match statement

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,406 @@
...
+		});
+	}
+
+	Ok(MangaPageResult {
+		manga: mangas,
+		has_more: false,
+	})
+}
+
+pub fn parse_manga_listing(base_url: String, url: String, _listing: Listing, _page: i32) -> Result<MangaPageResult> {
```
Skittyblock, line 128: you don't need the `listing` and `page` parameters if you aren't using them

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,406 @@
...
+
+	let data = json
+		.get("data")
+		.as_array()
+		.expect("Failed to get data as object");
+
+	let mut mangas: Vec<Manga> = Vec::new();
+
+	for manga in data {
+		let manga = manga.as_object().expect("Failed to get manga as object");
```
Skittyblock, line 144: looks like you're parsing it the same way as in `parse_manga_list`, so you could unify it and avoid less code duplication

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,406 @@
...
+				volume: volume as f32,
+				chapter: index.parse::<f32>().unwrap_or_default(),
+				date_updated: date_updated,
+				scanlator: String::new(),
+				url: url,
+				lang: String::from("en"),
+			});
+			
+			
+
```
Skittyblock, line 331: don't need extra blank lines
```suggestion
```

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,406 @@
...
+			if price != 0 {
+				continue;
+			}
+
+			let name = chapter.get("chapter_name").as_string().expect("Failed to get chapter name as str").read();
+			let index = chapter.get("index").as_string().expect("Failed to get chapter index as int").read();
+			let chapter_slug = chapter.get("chapter_slug").as_string().expect("Failed to get chapter slug as str").read();
+			let url = format!("{}/series/{}/{}", base_url, manga_id, chapter_slug);
+			let date_updated = chapter.get("created_at").as_string().expect("").read();
+			//aidoku::prelude::println!("{}", date_updated);
```
Skittyblock, line 316:
```suggestion
```

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,406 @@
...
+	}
+		
+
+	Ok(all_chapters)
+}
+
+pub fn parse_page_list(base_url: String, manga_id: String, chapter_id: String) -> Result<Vec<Page>> {
+	let url = format!("{}/series/{}/{}", base_url, manga_id, chapter_id);
+
+	let obj = Request::new(url.as_str(), HttpMethod::Get).html()?;
```
Skittyblock, line 342: is `.as_str()` necessary? if so, you could probably just use `&` instead which is simpler

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,406 @@
...
+
+	let mut page_list: Vec<Page> = Vec::new();
+	
+	for (i, page) in obj.select("img").array().enumerate() {
+		let obj = page.as_node().expect("node array");
+		let url = obj.attr("data-src").read();
+		let url2 = obj.attr("src").read();
```
Skittyblock, line 349: no point reading this if `url` exists

Skittyblock (changes requested):
- You can run `cargo +nightly clippy` to get some more things to fix, or suggestions for better ways of doing things
- Make sure to run `cargo +nightly fmt`
- Icon isn't fully opaque. Ideally, use a rectangular icon, but if that doesn't exist just add a background and resize to 128x128.

src/rust/en.omegascans/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[get_manga_listing]
+fn get_manga_listing(listing: Listing, _page: i32) -> Result<MangaPageResult> {
+	let list_query = match listing.name.as_str() {
+		"Latest" => "latest",
+		"Popular" => "total_views",
+		"Alphabetical" => "title",
+		_ => "",
+	};
+	let url = format!("https://api.omegascans.org/query?query_string=&series_status=All&order=desc&orderBy={}&series_type=Comic&page=1&perPage=1000&tags_ids=[]", list_query);
```
Skittyblock, line 26: you should define `BASE_API_URL` in this file and use it here as well

src/rust/en.omegascans/src/parser.rs
Skittyblock: this still -- why are you using `push_str`?

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,233 @@
...
+	if !genres.is_empty() {
+		genres.pop();
+	}
+
+	let url = format!("{}/query?query_string={}&series_status={}&order=desc&orderBy=total_views&series_type=Comic&page=1&perPage=1000&tags_ids=[{}]", BASE_API_URL, search_query, status, genres);
+	let json = Request::new(url, HttpMethod::Get);
+	let mangas = parse_manga(base_url.clone(), json)?;
+
+	Ok(MangaPageResult {
+		manga: mangas,
```
Skittyblock, line 57: you should just call the variable manga and then you can do this (plural of manga is manga)
```suggestion
	let manga = parse_manga(base_url.clone(), json)?;

	Ok(MangaPageResult {
		manga,
```

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,233 @@
...
+		has_more: false,
+	})
+}
+
+pub fn parse_manga_details(base_url: String, manga_id: String) -> Result<Manga> {
```
Skittyblock, line 72: neither of these need to take ownership, but particularly the `base_url`
```suggestion
pub fn parse_manga_details(base_url: &String, manga_id: String) -> Result<Manga> {
```

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,233 @@
...
+
+	Ok(MangaPageResult {
+		manga: mangas,
+		has_more: false,
+	})
+}
+
+pub fn parse_manga_listing(base_url: String, url: String) -> Result<MangaPageResult> {
+	let json = Request::new(url, HttpMethod::Get);
+	let mangas = parse_manga(base_url.clone(), json)?;
```
Skittyblock, line 64: don't `clone()` if you can help it! pass a reference, there's no reason the function needs to take ownership of the string when you're just using it to construct a new one
```suggestion
	let manga = parse_manga(&base_url, json)?;
```

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,233 @@
...
+
+	Ok(Manga {
+		id,
+		cover,
+		title,
+		author,
+		artist,
+		description,
+		url,
+		categories: Vec::new(),
```
Skittyblock, line 104: use `..Default::default()` here too
```suggestion
```

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,233 @@
...
+		let url = format!("{}/series/{}", base_url, id);
+
+		mangas.push(Manga {
+			id,
+			cover,
+			title,
+			url,
+			status: manga_status,
+			nsfw: MangaContentRating::Nsfw,
+			viewer: MangaViewer::Scroll,
```
Skittyblock, line 227: you don't need all this data for the listing; just id, cover, title, and url is good. save the rest for `parse_manga_details`.
```suggestion
```

Skittyblock (changes requested):
just a few more changes

Aeonss:
build.sh should have the right permissions (-rwxrwxr-x), but is still failing to build.

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,213 @@
+use aidoku::{
+	error::Result, helpers::uri::encode_uri_component, prelude::format, std::net::HttpMethod,
+	std::net::Request, std::String, std::Vec, Chapter, Filter, FilterType, Manga,
+	MangaContentRating, MangaPageResult, MangaStatus, MangaViewer, Page,
+};
+
+const BASE_API_URL: &str = "https://api.omegascans.org";
```
Skittyblock, line 7: this can be imported from `lib.rs` instead (I think `crate::BASE_API_URL` would work).

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,213 @@
...
+			},
+			_ => continue,
+		}
+	}
+
+	if !genres.is_empty() {
+		genres.pop();
+	}
+
+	let url = format!("{}/query?query_string={}&series_status={}&order=desc&orderBy=total_views&series_type=Comic&page=1&perPage=1000&tags_ids=[{}]", BASE_API_URL, search_query, status, genres);
```
Skittyblock, line 54: wait I just realized, why on earth are you loading 1000 manga in a single page? use the `page` parameter and do like 20 per page instead, same goes for listing

Aeonss: Oh sorry, that was previously before I knew about the page thing. I did it for webtoonxyz but not for omegascans lol

src/rust/en.omegascans/src/parser.rs
```diff
@@ -0,0 +1,226 @@
...
+			..Default::default()
+		});
+	}
+
+	Ok(mangas)
+}
+
+fn is_last_page(url: &String) -> bool {
+	let json = Request::new(url, HttpMethod::Get);
+	!json.json().expect("").as_object().expect("").get("data").as_array().expect("").is_empty()
```
Skittyblock, line 225: - you're re-requesting the page even when you already have a json from it (probably just check if `manga` is empty)
- don't use `expect` unless you are certain it will never be called. this will break the source until the app is restarted
- make sure there's a newline at the end of the file

Aeonss: Oops, sorry about that. Forgot I could just check the manga array. I used expect since it said I couldn't use ? in a bool function, but that's unnecessary anyway. Hope this is the last fix, thanks for the code reviews.

Skittyblock (approved):
in the future, you don't need to have a base_url for all the functions if it's not a template source, but it's fine to leave it for now I think

PR #612, opened 2024-03-18
feat: add source mangamonks

daarkdemon:
Checklist:


src/rust/en.mangamonks/src/helper.rs
```diff
@@ -0,0 +1,88 @@
...
+	excluded_tags: Vec<String>,
+	manga_type: String,
+	status: String,
+	page: i32,
+) -> String {
+	let mut url = format!("{}/genre/action/{}?term=", base_url, page);
+	if !included_tags.is_empty() || !excluded_tags.is_empty() {
+		if excluded_tags.is_empty() {
+			for tag in included_tags {
+				url.push_str(&format!("&include%5B%5D={}", tag));
```
Skittyblock, line 66: in the future it would probably be cleaner to use `QueryParameters`

src/rust/en.mangamonks/src/parser.rs
```diff
@@ -0,0 +1,149 @@
...
+		..Default::default()
+	})
+}
+
+pub fn parse_chapter_list(html: Node) -> Result<Vec<Chapter>> {
+	let mut chapters: Vec<Chapter> = Vec::new();
+
+	for chapter in html.select(".chapter-list li").array() {
+		let node = chapter.as_node().expect("Failed to get node");
+		let title = node.select("span:nth-child(1)").text().read();
```
Skittyblock, line 118: if the chapter title is always "Chapter X" (seems to be), you shouldn't include the title since Aidoku will automatically format it

src/rust/en.mangamonks/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+			.header(
+				"Content-Type",
+				"application/x-www-form-urlencoded; charset=UTF-8",
+			)
+			.header("Accept", "application/json")
+			.header("X-Requested-With", "XMLHttpRequest")
+			.body(parameters)
+			.json()?
+			.as_object()?;
+		let results = json.get("manga").as_array()?;
```
Skittyblock, line 85: I think this should be a function in `parser.rs` since you're parsing the manga list

src/rust/en.mangamonks/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+	let url = get_search_url(
+		BASE_URL,
+		included_tags,
+		excluded_tags,
+		manga_type,
+		status,
+		page,
+	);
+	let html = Request::new(url, HttpMethod::Get).html()?;
+	parse_manga_listing(html)
```
Skittyblock, line 129: no need for an explicit `return`, you can just have an else case and use implicit returns
```suggestion
		Ok(MangaPageResult {
			manga: mangas,
			has_more: false,
		})
	} else {
		let url = get_search_url(
			BASE_URL,
			included_tags,
			excluded_tags,
			manga_type,
			status,
			page,
		);
		let html = Request::new(url, HttpMethod::Get).html()?;
		parse_manga_listing(html)
	}
```

Skittyblock (changes requested):

Skittyblock (approved):

PR #634, opened 2024-04-13
feat: add libgroup template and mangalib source

vlapsk1y:
Checklist:

I know about this pq (#176), but it's not merged into main branch and soon stay deprecated, because mangalib.me in the near future get very major update (test-front.mangalib.me). The new version of mangalib has become unparseable and open at browser as web app.


src/rust/multi.libgroup/template/src/parser.rs
```diff
@@ -0,0 +1,171 @@
...
+                title: title,
+                author: String::new(),
+                artist: String::new(),
+                description: String::new(),
+                url: url,
+                categories: Vec::new(),
+                status: status,
+                nsfw: MangaContentRating::Safe,
+                viewer: aidoku::MangaViewer::Rtl
+            })
```
Skittyblock, line 54: use `..Default::default` at the end of the struct and exclude the fields you don't populate
```suggestion
            arr_manga.push(Manga {
                id,
                cover,
                title,
                url,
                status,
                nsfw: MangaContentRating::Safe,
                viewer: aidoku::MangaViewer::Rtl,
                ..Default::default
            })
```

Skittyblock (changes requested):
- jsons should use tabs for indentation 
- files should have trailing newlines
- run `cargo +nightly fmt`
- I'm not sure about the "site id" stuff, especially since you're only adding mangalib. the template struct should probably just provide the url and route and stuff, with the actual sources containing the stuff rather than the template hardcoding it.

vlapsk1y:
as for "site id", I wanted to make the task easier for those who want to implement other sites of libgroup. Although them still need to edit template, cuz nsfw sites has own cdn. I really didn't knew about that fact, until checked before writing the comment.

Skittyblock:
yes but the point of the template is to contain all things specific to the template that the individual websites use, but then it's the job of the actual sources to just use the template while providing the source specific things. for example, the individual source should be able to give any base url to the template, and the template shouldn't need to know which exact source it is.

vlapsk1y:
I understood your point, I'll make a commit in the next days. thanks u.

src/rust/multi.libgroup/sources/mangalib/res/filters.json
Skittyblock, line 1: jsons are still using spaces for indentation instead of tabs

Skittyblock:
icon should be 128x128 and make sure to fix the warnings clippy generates when running `cargo +nightly clippy`, but after that it should be good

src/rust/multi.libgroup/sources/mangalib/res/filters.json
Skittyblock, line 1: ids need to be strings

Skittyblock (approved):

PR #635, opened 2024-04-14
fix(madara): use alternate method to get listings

### LINKED ISSUE #632: ManhuaPlus not working
Not sure if it has anything to do with the website's problems (as stated by manhuaplus) or not but the source doesn't load.  If it isn't possible to fix this source, is it possible to transfer/replace the source with manhuaplus.org? (Original site is manhuaplus.com)

Thank you for any feedback on this.

KaBankz:
Quick and dirty fix for some madara sources.

Since most sites work without this, it's not enabled by default, but it should work with all madara sources regardless, but I ain't taking that risk, it can be toggled on a case by case basis.

Closes #632

Checklist:

### DIFF
--- a/src/rust/madara/template/src/template.rs
+++ b/src/rust/madara/template/src/template.rs
@@ -270,8 +270,8 @@ pub fn get_series_page(data: MadaraSiteData, listing: &str, page: i32) -> Result
 		};
 
 		let url = format!(
-			"https://manhuaplus.com/page/{}?s&post_type=wp-manga&m_orderby={}",
-			page, listing
+			"{}/page/{}?s&post_type=wp-manga&m_orderby={}",
+			data.base_url, page, listing
 		);
 
 		return get_search_result(data, url);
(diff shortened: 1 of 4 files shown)


Skittyblock (dismissed):

src/rust/madara/template/src/template.rs
```diff
@@ -257,6 +260,23 @@ pub fn get_search_result(data: MadaraSiteData, url: String) -> Result<MangaPageR
...
+	if !data.use_ajax_listing {
+		let listing = match listing {
+			"_wp_manga_views" => "views",
+			"_wp_manga_week_views_value" => "trending",
+			"_latest_update" => "latest",
+			_ => "latest",
+		};
+
+		let url = format!(
+			"https://manhuaplus.com/page/{}?s&post_type=wp-manga&m_orderby={}",
```
Skittyblock, line 273: wait, shouldn't this be using the template base url, not just manhuaplus?

KaBankz: oh i completely missed that

Skittyblock (approved):

PR #641, opened 2024-04-19
feat: manga demon

KaBankz:
🎉

Checklist:


Skittyblock (changes requested):
just make sure to change the `expect`s to `?`s on the requests, since if the user opens the source when not connected to the internet it'd just break

Skittyblock (approved):

KaBankz:
Looks like they changed their domain within the last week, I'll pr an update soon™

PR #647, opened 2024-04-25
feat(zeroscans)!: rewrite

KaBankz:
> [!CAUTION]
> The comic id's logic has changed so migration will be required

Rewrote most of the logic to cache all the comics and use said cache to avoid making extraneous requests when possible.

Filters and deeplinking were also implemented

Checklist:


Skittyblock (approved):
not doing any pagination seems a little iffy, but it's probably fine since there aren't that many total titles

KaBankz:
> not doing any pagination seems a little iffy, but it's probably fine since there aren't that many total titles

Their comics api endpoint doesn't have any pagination, it returns all the comics in a single response.

Skittyblock:
Yeah I know, but like with MangaSee we paginate it ourselves for aidoku since there's no sense sending all of the data to it if the user doesn't scroll down

KaBankz:
> we paginate it ourselves for aidoku

Oh, didn't know that

> but it's probably fine since there aren't that many total titles

Yeah, they have like 100 titles total, so not a big issue.

PR #661, opened 2024-05-11
fix(mangadex): do not fail chapter parsing if `externalUrl` is null

hanatsumi:
Looking at #25, it looks like the idea was to filter out chapters where the `externalUrl` is not null; however the current logic filters both when `externalUrl` is `null` and when it's a string (so essentially always I think?). Changed the logic to filter only chapters where `externalUrl` is a string.

A manga where chapters are filtered incorrectly is Houseki no Kuni ([API call](https://api.mangadex.org/manga/cade38b7-64c4-4a29-8e3c-8c283291d6c6/feed?order[volume]=desc&order[chapter]=desc&limit=500&contentRating[]=pornographic&contentRating[]=erotica&contentRating[]=suggestive&contentRating[]=safe&includes[]=scanlation_group&translatedLanguage[]=en), note how `externalUrl` is always null).

I haven't tested this with Aidoku itself (as I do not own a iOS/macOS device), but with a custom client that uses Aidoku's sources; it seems to work properly

Checklist:

### DIFF
--- a/src/rust/multi.mangadex/res/source.json
+++ b/src/rust/multi.mangadex/res/source.json
@@ -4,7 +4,7 @@
 		"lang": "multi",
 		"name": "MangaDex",
 		"url": "https://mangadex.org",
-		"version": 7,
+		"version": 8,
 		"minAppVersion": "0.5"
 	},
 	"languages": [
--- a/src/rust/multi.mangadex/src/parser.rs
+++ b/src/rust/multi.mangadex/src/parser.rs
@@ -240,8 +240,7 @@ pub fn parse_chapter(chapter_object: ObjectRef) -> Result<Chapter> {
 
 	// Fix for Skittyblock/aidoku-community-sources#25
 	let ext_url = attributes.get("externalUrl");
-	if ext_url.is_none()
-		|| ext_url.as_string().is_ok()
+	if ext_url.as_string().is_ok()
 		|| date_updated > crate::helper::current_date()
 	{
 		return Err(aidoku::error::AidokuError {


Skittyblock:
Honestly, looking at the code, it seems like you're right, but in practice these chapters aren't actually being filtered out. If you look at Houseki no Kuni in Aidoku, all the proper chapters are there. @beerpiss do you know why the code is checking if the fetched attribute is none?

hanatsumi:
Yeah, it's a bit weird to me that this works within Aidoku, because by looking at [`ValueRef`'s implementation](https://github.com/Aidoku/aidoku-rs/blob/004bddabade7b24c58cf925b08f90dd093b00c9d/crates/imports/src/std.rs#L124) it seems like either the check for a `ValueKind::Null` (inside `is_none()`) or for a `ValueKind::String` (inside `as_string()`) will succeed; so I'd assume `ext_url` is some other `ValueKind` here?

beerpiss:
> Honestly, looking at the code, it seems like you're right, but in practice these chapters aren't actually being filtered out. If you look at Houseki no Kuni in Aidoku, all the proper chapters are there. @beerpiss do you know why the code is checking if the fetched attribute is none?

It's been a super long time that I couldn't remember, sorry. Maybe I copied it from the Tachi extension?

Skittyblock:
@hanatsumi do you know any titles that actually come out differently with the change? I think I'm probably fine with just merging it but I'm curious if there's an actual difference

hanatsumi:
@Skittyblock On my [custom client](https://github.com/hanatsumi/rakuyomi), pretty much all titles change, because as it is now chapter parsing always fails due to the `is_none` check. I don't really know if any titles change within Aidoku, but I believe the behaviour should stay the same.

Skittyblock:
Alright, likely an issue with Aidoku's api or aidoku-rs then.

Skittyblock (approved):

PR #685, opened 2024-07-04
Mangademon Update

AlokSwaminathan:
Checklist:
- [ x] Updated source's version for individual source changes
- [ x] Updated all sources' versions for template changes
- [ x] Set appropriate `nsfw` value
- [ x] Did not change `id` even if a source's name or language were changed
- [ x] Tested the modifications by running it on the simulator or a test device 

Previously the source would load pictures and manga titles but would not load chapters or descriptions, just a small error in the URL params it looks like because with my small updates it works perfectly again on my iPhone 15 running iOS 17.5.1


src/rust/en.mangademon/res/source.json
```diff
@@ -3,8 +3,8 @@
 		"id": "en.mangademon",
 		"lang": "en",
 		"name": "Manga Demon",
-		"version": 2,
-		"url": "https://comicdemons.com",
+		"version": 3,
+		"url": "https://mgdemon.org/",
```
Skittyblock, line 7: there shouldn't be a trailing slash
```suggestion
		"url": "https://mgdemon.org",
```

src/rust/en.mangademon/src/lib.rs
```diff
@@ -17,7 +17,7 @@ use aidoku::{
 use helper::*;
 use parser::*;
 
-const BASE_URL: &str = "https://comicdemons.com";
+const BASE_URL: &str = "https://mgdemon.org/";
```
Skittyblock, line 20:
```suggestion
const BASE_URL: &str = "https://mgdemon.org";
```

src/rust/en.mangademon/src/lib.rs
```diff
@@ -56,10 +56,10 @@ fn get_manga_list(filters: Vec<Filter>, page: i32) -> Result<MangaPageResult> {
...
 				let ascending = value.get("ascending").as_bool().unwrap_or(false);
 				url.push_str(match (index, ascending) {
-					(0, true) => "&orderby=name%20asc",
-					(0, false) => "&orderby=name%20desc",
-					(1, true) => "&orderby=views%20asc",
-					(1, false) => "&orderby=views%20desc",
+					(0, true) => "&orderby=NAME%20ASC",
+					(0, false) => "&orderby=NAME%20ASC",
+					(1, true) => "&orderby=VIEWS%20ASC",
+					(1, false) => "&orderby=VIEWS%20ASC",
```
Skittyblock, line 62: why did you change the capitalization here? it seems like it works fine either way on the site

Skittyblock (changes requested):
make sure to pull main now that aidoku-rs was updated so it fixes the workflow

src/rust/en.mangademon/src/lib.rs
AlokSwaminathan: It is just what the site uses by default for getting new pages, so I thought it would be best to replicate that.

KaBankz: You replaced the descending values with ascending btw

Skittyblock (approved):

PR #690, opened 2024-07-12
Omega Scans: Added a 'newest' listing and fixed missing last image

Aeonss:
Checklist:


src/rust/en.omegascans/res/source.json
```diff
@@ -3,17 +3,20 @@
 		"id": "en.omegascans",
 		"lang": "en",
 		"name": "Omegascans",
```
Skittyblock, line 5: looks like this is how they have it on the site
```suggestion
		"name": "Omega Scans",
```

Skittyblock (approved):

PR #695, opened 2024-07-16
Add Faccina

LetrixZ:
Checklist:


src/rust/multi.faccina/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+		})
+}
+
+#[get_chapter_list]
+fn get_chapter_list(id: String) -> Result<Vec<Chapter>> {
+	let base_url = get_base_url()?;
+	let url = format!("{base_url}/g/{id}");
+	Ok(vec![Chapter {
+		id,
+		title: "Chapter".to_string(),
```
Skittyblock, line 110: I think it's better to leave the title empty so that aidoku just automatically sets the title as "Chapter 1" (rather than "Ch. 1 - Chapter")

Skittyblock (changes requested):
code looks good to me, just a few changes:
- json indentation should be tabs
- icon needs a background color (you could just make it white)
- make sure all files have a newline at the end (settings.json, cargo.toml)
- aidoku-rs is broken right now (sorry), but once it's updated you'll need to update it on your end and then pull main to fix the building for the rest of the sources

LetrixZ:
I don't have an iPhone anymore so this isn't a priority. And I'm also doing a rewrite of Faccina.
I'll see if I revisit this later when it's done.

PR #756, opened 2024-09-04
fix(mangademon): rewrite most of the source to account for a site remake

### LINKED ISSUE #753: Manga Demon app source not working
I’m not sure what else to say but the website works just fine but something is wrong with the source in app.

AlokSwaminathan:
Checklist:

Closes #753

### DIFF
--- a/src/rust/en.mangademon/src/parser.rs
+++ b/src/rust/en.mangademon/src/parser.rs
@@ -186,7 +186,7 @@ pub fn parse_chapter_list(html: Node) -> Vec<Chapter> {
 		let num_str = String::from(num_str.split(" ").last().unwrap_or("1"));
 		let chapter = num_str.parse::<f32>().unwrap_or(-1.0);
 
-    let id = get_chapter_id(&url);
+		let id = get_chapter_id(&url);
 
 		let lang = String::from("en");
 
@@ -206,15 +206,15 @@ pub fn parse_chapter_list(html: Node) -> Vec<Chapter> {
 }
 
 pub fn parse_page_list(html: Node) -> Vec<Page> {
-  let mut pages = Vec::new();
+	let mut pages = Vec::new();
 
-  for (index, node) in html.select(".imgholder").array().enumerate() {
+	for (index, node) in html.select(".imgholder").array().enumerate() {
 		let url = node
 			.as_node()
 			.expect("Failed to get chapter image")
 			.attr("abs:src")
 			.read();
-    let index: i32 = index.try_into().unwrap();
+		let index: i32 = index.try_into().unwrap();
 		pages.push(Page {
 			index,
 			url,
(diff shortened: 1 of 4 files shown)


src/rust/en.mangademon/src/parser.rs
```diff
@@ -61,122 +98,104 @@ pub fn parse_manga_list(html: Node, searching: bool) -> MangaPageResult {
...
-				rating = MangaContentRating::Suggestive;
-			}
-			if categories.iter().any(|e| e == "Smut") {
-				rating = MangaContentRating::Nsfw;
+		for genre in categories.iter() {
+			match genre.to_lowercase().trim() {
+				"Ecchi" | "Harem" | "Mature" => rating = MangaContentRating::Suggestive,
+				"Smut" => rating = MangaContentRating::Nsfw,
+				_ => {}
 			}
```
Skittyblock, line 149: you're matching a lowercase string against stuff with capitals, so it'll always fail

also should probably `break` after setting the nsfw rating, because it could loop again and set the rating to suggestive instead of nsfw like it should be
```suggestion
			match genre.to_lowercase().trim() {
				"ecchi" | "harem" | "mature" => rating = MangaContentRating::Suggestive,
				"smut" => {
					rating = MangaContentRating::Nsfw;
					break;
				}
				_ => {}
			}
```

Skittyblock (changes requested):

src/rust/en.mangademon/src/parser.rs
```diff
@@ -61,122 +98,107 @@ pub fn parse_manga_list(html: Node, searching: bool) -> MangaPageResult {
...
 		categories,
+		description,
+		author,
+		artist,
 		status,
 		nsfw,
+		cover,
 		viewer,
+		url: manga_url,
 		..Default::default()
```
Skittyblock, line 174: there's a clippy warning here since this isn't necessary
```suggestion
```

Skittyblock (changes requested):
seeing a crash when opening martial peak due to duplicate chapter identifiers. double check how you're assigning chapter ids and make sure they're unique, and make sure to test on device

AlokSwaminathan:
> seeing a crash when opening martial peak due to duplicate chapter identifiers. double check how you're assigning chapter ids and make sure they're unique, and make sure to test on device

I've been reading from mangademon with this version of the source for a few weeks, I have 60 from it and I haven't noticed any issues. What chapter on martial peak are you having problems with, I tested a few chapters on there and they seem to work for me.

Skittyblock:
interestingly enough, it's only happening when run on the simulator. none of the images load for me on device though. the culprit for martial peak is `/chaptered.php?manga=1&chapter=3705`, which I can see on the site is listed twice. this is their mistake, but you should probably filter out duplicates on our end within the source

AlokSwaminathan:
I took a look at 3705 and it seems to work on my phone even though there were chapters with that number. I just went ahead and removed duplicates anyway though

Skittyblock (approved):

PR #757, opened 2024-09-09
feat(GenZToons): Removing SuryaToons - Adding GenZToons

Kareylo:
As Surya Toons is now dead, I removed it.
The new Team name of Surya Toons is GenZ Toons.

Still the same template,

Checklist:


src/rust/mangastream/sources/genztoons/res/source.json
```diff
@@ -1,10 +1,10 @@
...
 	"info": {
-		"id": "en.suryatoon",
+		"id": "en.genz",
 		"lang": "en",
-		"name": "Surya Toon",
-		"version": 2,
-		"url": "https://suryatoon.com/",
+		"name": "GenZ Toon",
+		"version": 1,
+		"url": "https://genztoons.com/",
```
Skittyblock, line 7: base url shouldn't have a trailing slash -- didn't catch this for suryatoon I guess
```suggestion
		"url": "https://genztoons.com",
```

Skittyblock (changes requested):
do the sources have the same content and ids? if so, it would be better to keep the old id and just change the name and bump the version. if not though, this is fine

Kareylo:
> do the sources have the same content and ids? if so, it would be better to keep the old id and just change the name and bump the version. if not though, this is fine

It seems it's exactly same content and ids, I switched back to old source id and bumped the version

Skittyblock (approved):

PR #782, opened 2024-10-20
fix(komga): Non-English search not working

### LINKED ISSUE #659: Komga source not support Non-English search
While using the Komga source on aidoku I found that it currently only supports alphanumeric to search for content, not non-English searches such as Chinese.

dyphire:
## Checklist:

## Related Issues
Close #659
## Screenshots

| Before |	After |
| ------- | ------ |
| image | image |

### DIFF
--- a/src/rust/multi.komga/res/source.json
+++ b/src/rust/multi.komga/res/source.json
@@ -3,7 +3,7 @@
 		"id": "multi.komga",
 		"lang": "multi",
 		"name": "Komga",
-		"version": 2,
+		"version": 3,
 		"url": "https://komga.org",
 		"nsfw": 0
 	},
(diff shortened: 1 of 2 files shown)


Skittyblock (changes requested):
make sure you fix the clippy warning. 

this still works for searches containing special characters like & and spaces, correct?

dyphire:
> make sure you fix the clippy warning.

Changed as suggested by clippy.

> this still works for searches containing special characters like & and spaces, correct?

Searching for special characters has never worked correctly because komga doesn't do url decoding on the search response. That's why non-English searches didn't work before, because aidoku encodes the url and komga doesn't decode the url accordingly. This is a limitation of komga.

dyphire:
> Any more questions here?

Here, too.

Skittyblock (approved):

PR #785, opened 2024-10-21
fix(manhuagui): incorrect chapter order

dyphire:
Checklist:

### DIFF
--- a/src/rust/zh.manhuagui/res/source.json
+++ b/src/rust/zh.manhuagui/res/source.json
@@ -3,8 +3,8 @@
 		"id": "zh.manhuagui",
 		"lang": "zh",
 		"name": "ManHuaGui",
-		"version": 2,
-		"url": "https://www.manhuagui.com/",
+		"version": 3,
+		"url": "https://www.manhuagui.com",
 		"nsfw": 1
 	}
 }
--- a/src/rust/zh.manhuagui/src/decoder.rs
+++ b/src/rust/zh.manhuagui/src/decoder.rs
@@ -36,8 +36,7 @@ impl Decoder {
 			String::new()
 		};
 
-		let suffix_vec = vec![
-			self.tr(c % self.a, 36),
+		let suffix_vec = [self.tr(c % self.a, 36),
 			String::from_utf8(vec![(c % self.a + 29) as u8]).unwrap_or_default(),
 		];
 		let suffix = suffix_vec[(c % self.a > 35) as usize].clone();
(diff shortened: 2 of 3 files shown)


src/rust/zh.manhuagui/src/decoder.rs
```diff
@@ -36,8 +36,7 @@ impl Decoder {
 			String::new()
 		};
 
-		let suffix_vec = vec![
-			self.tr(c % self.a, 36),
+		let suffix_vec = [self.tr(c % self.a, 36),
```
Skittyblock, line 39: why was the formatting here changed? it was better before

src/rust/zh.manhuagui/res/source.json
```diff
@@ -3,7 +3,7 @@
 		"id": "zh.manhuagui",
 		"lang": "zh",
 		"name": "ManHuaGui",
-		"version": 2,
+		"version": 3,
 		"url": "https://www.manhuagui.com/",
```
Skittyblock, line 7: should change this while we're at it
```suggestion
		"url": "https://www.manhuagui.com",
```

Skittyblock (changes requested):
does this really change the chapter ordering? you're reversing the way you parse all the elements, but then you reverse the whole chapter list again, so doesn't it end up being the same? do you have an example manga that has incorrect ordering with the old version but is fixed with this version?

src/rust/zh.manhuagui/src/decoder.rs
dyphire: clippy would complain about it when testing in my repository, so fix that.

dyphire:
> does this really change the chapter ordering? you're reversing the way you parse all the elements, but then you reverse the whole chapter list again, so doesn't it end up being the same? do you have an example manga that has incorrect ordering with the old version but is fixed with this version?

This is not true, it is done because of the special structure of the chapter list in this web page, which was overlooked by the previous people who implemented this source. 
It can be confirmed by checking the web page: https://www.manhuagui.com/comic/28004/

manhuagui v2: 
IMG_4931
This PR:
IMG_4933
IMG_4932

src/rust/zh.manhuagui/res/source.json
dyphire: My computer has been turned off and can't be handled for the time being.

dyphire: done

dyphire:
Any more questions here?

Skittyblock (approved):
alright, if you say it's correct then sure

PR #804, opened 2024-11-19
fix(asurascans): fix chapter names and page list parsing

AlokSwaminathan:
Checklist:

### DIFF
--- a/src/rust/en.asurascans/src/lib.rs
+++ b/src/rust/en.asurascans/src/lib.rs
@@ -325,7 +325,7 @@ fn get_page_list(manga_id: String, chapter_id: String) -> Result<Vec<Page>> {
       if pages.iter().any(|page| page.index == index) {
         continue;
       }
-      println!("index: {}, url: {}", index, url);
+
       pages.push(Page {
         index,
         url,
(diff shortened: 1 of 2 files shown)


src/rust/en.asurascans/src/lib.rs
```diff
@@ -299,31 +299,41 @@ fn get_chapter_list(manga_id: String) -> Result<Vec<Chapter>> {
...
-			index,
-			url,
-			..Default::default()
-		});
+  let mut text_slice = html_text.as_str();
+  loop {
+    let chap = text_slice.find("https://gg.asuracomic.net/storage/media/");
+    if let Some(chap) = chap {
+      text_slice = &text_slice[chap..];
+      let end = text_slice.find(".webp").unwrap();
```
Skittyblock, line 311: don't use `unwrap`, if it doesn't exist either continue or return early with an error

Skittyblock (changes requested):
make sure clippy passes; i think it will give you some error about the indentation in lib.rs

edit: i guess clippy passed in the workflow? make sure the file is properly indented though

AlokSwaminathan:
I had some space based tabbing on accident, I think that's fixed now.

Skittyblock (approved):
i guess we can have this as a temporary fix

PR #838, opened 2024-12-30
fix(nhentai): new image subdomain

axiel7:
Checklist:

### DIFF
--- a/src/rust/multi.nhentai/res/source.json
+++ b/src/rust/multi.nhentai/res/source.json
@@ -3,7 +3,7 @@
 		"id": "multi.nhentai",
 		"lang": "multi",
 		"name": "NHentai",
-		"version": 7,
+		"version": 8,
 		"url": "https://nhentai.net/",
 		"nsfw": 2
 	},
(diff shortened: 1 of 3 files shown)


shiroyasha9:
Bumping this PR, if it needs any changes, I can contribute

axiel7:
This already works but only with new manga, old entries still use the i1 image server so it's broken for those.

We need to get the correct image server for each item. This info is on one of the html `<script>` section (see [mihon extension](https://github.com/keiyoushi/extensions-source/commit/bc7e4625164a4a5ea18c26300f34dae2a1570f1d#diff-309762db333311f74f0437432527f632d4ff1ec8eeff59f2421c388931e1a352R264)).

My knowledge in rust is very limited so any help is appreciated

shiroyasha9:
Because we're not scraping the page and using the /api directly, I'm pretty sure we won't be able to get the `media_server` directly (as the mihon extension is doing)

shiroyasha9:
How much old mangas are talking about here? possibly there could be 2 versions of this extension - one for old and one for new?

axiel7:
Yeah since we're using the API we need to make an extra request to get the html content page

axiel7:
I just tested the merged changes and works fine for me. You tried to load the site first from the 🧭 icon?

PR #840, opened 2025-01-03
feat(libgroup): implementation of login to account SocialLib

### LINKED ISSUE #805: Mangalib missing chapters
When I’m using a Mangalib source, some titles can’t be opened by aidoku. It says “no chapters”, but on the website I can read it
For example Vinland Saga 
I’m a russian speaking person, so I want to read manga in russian. All of the multilanguage sources are out of date for me

vlapskii:
This feature gives access to NSFW libgroup sites and access to manga chapters tagged 18+, which are not available without an account (closes #805)

Checklist:


maksim-katafalk:
When would it be fixed?

vlapskii:
I'm spending the New Year holidays with my family and my laptop is not always at hand. I'm currently writing code to parse web pages for authorization.

maksim-katafalk:
Btw thanks for the work

vlapskii:
Well... I can't make this function until aidoku-rs doesn't have a method to disable request redirect :skull:

SocialLib in endpoint auth/login giving a response with a redirect on HTTP, and it seems breaks server side, because it redirects on first endpoint when it must redirect on page with hidden tokens in html code. Recent browsers use https protocol by default and don't notice http, they just replace http with https in URL when aidoku-rs redirects directly.

maksim-katafalk:
Well, it turns out we have to wait until the Aidoku developer fixes this bug. And as I understand it, it can take a long time.

vlapskii:
that's not bug

maksim-katafalk:
but what is needed to solve this problem?

maksim-katafalk:
I see, so we'll have to wait until the method to disable request redirect appears? And it's just a matter of time until they add this feature?

vlapskii:
Yes. We have to wait until the method is added to the API, but as far as I know, a new Aidoku API (0.7) will appear soon, which will have this method, and I have rewrite source for the API.

Ну да по сути надо ждать пока не появится метод в API, но вроде бы скоро будет совершенно новый API, поэтому мне придется переписывать расширение, оно работать-то будет, но просто будет с пометкой легаси, пока не перепишу под новый. Кстати, о языке, здесь я буду исключительно на английском коммуницировать чисто из правила хорошего тона, это соообщение — исключение.

maksim-katafalk:
Got it, all that's left is to wait for the new version of the app to come out. Thanks for the info, I'm looking forward to you fixing this.

PR #865, opened 2025-01-30
fix(nhentai): image loading

### LINKED ISSUE #638: NHentai Popular : All-Time sorting not loading
Other sorting loads fine

### LINKED ISSUE #828: Nhentai images/cover do not load
Since no one reposted the issue regarding this based on [PinkBlossom's orginal issue](https://github.com/Skittyblock/aidoku-community-sources/issues/794)

Cover and pages no longer load since they are based on a new image format.

kkantan:
## Checklist

## Changes
switched to just parsing the html instead of using the api (for everything besides chapter list), because it seems like there's no way to get the proper media server for images from the api.

closes #828 
closes #638


src/rust/multi.nhentai/src/lib.rs
```diff
@@ -265,44 +213,46 @@ fn get_manga_details(id: String) -> Result<Manga> {
...
 }
 
 #[get_page_list]
 fn get_page_list(_: String, id: String) -> Result<Vec<Page>> {
-	let request = Request::new(helper::get_details_url(id).as_str(), HttpMethod::Get)
-		.header("User-Agent", USER_AGENT);
-	let json = request.json()?.as_object()?;
+	let html = Request::new(format!("{BASE_URL}/g/{id}/"), HttpMethod::Get)
+		.header("User-Agent", USER_AGENT)
+		.string()?;
```
Skittyblock, line 249: might be better to html this and maybe try and select the script tag, or around it, instead of loading the entire page into wasm memory, but probably fine

Skittyblock (approved):

kkantan:
yes, I just checked again and everything appears to function fine

kkantan:
sounds like something that would be better implemented on the app side. from what i know, many sources have issues with cloudflare. maybe clearing your network cache might help.

PR #870, opened 2025-01-30
fix(hentaifox): older image and page loading

kkantan:
## Checklist

## Changes
previous version didn't account for older page images which still used jpeg. this one actually fetches the proper image extension. also fixed some issues with the urls being incorrect and next page loading not working.

### DIFF
--- a/src/rust/en.hentaifox/src/lib.rs
+++ b/src/rust/en.hentaifox/src/lib.rs
@@ -3,6 +3,7 @@
 mod helper;
 
 extern crate alloc;
+use alloc::string::ToString;
 
 use aidoku::{
 	error::Result,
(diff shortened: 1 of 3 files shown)


src/rust/en.hentaifox/src/lib.rs
```diff
@@ -193,6 +156,27 @@ fn get_page_list(_manga_id: String, chapter_id: String) -> Result<Vec<Page>> {
...
+		} else {
+			None
+		}
+	} else {
+		None
+	};
+
+	let json = json
+		.and_then(|json| aidoku::std::json::parse(json).ok())
+		.and_then(|json| json.as_object().ok());
```
Skittyblock, line 178: could probably put this in the first json variable so you don't need to reassign

Skittyblock (approved):

PR #978, opened 2025-05-19
Update url hencube

dongtho123:
Checklist:


Skittyblock (changes requested):
the url in `source.json` also needs to be changed

Skittyblock:
the aix file is how aidoku sources are packaged. you can install it by opening it in the aidoku app.

however, it seems like you don't have a proper dev environment set up. in order to develop sources, you should be able to modify and compile them locally to test changes before making a pr. if you manage to fix the source and ensure it's working, feel free to open another pr.

PR #1015, opened 2025-07-21
Added it.lupiteam source

roguebankrupt:
Checklist:
- [ X] Updated source's version for individual source changes
- [ X] Updated all sources' versions for template changes
- [ X] Set appropriate `nsfw` value
- [ X] Did not change `id` even if a source's name or language were changed
- [ X] Tested the modifications by running it on the simulator or a test device


Skittyblock (changes requested):
it looks like you've committed a number of files that don't need to be here: everything in `target/`, the `aix` and `zip` files, the `exe` file, etc.

also, this repo is going to be unmaintained going forward, and the sources here all use the *old* app api. it looks like you made a new source with the old api. if you clean this pr up i don't really mind merging it, but it would be better if you wrote the source using the new api and pr'd it to [here](https://github.com/Aidoku-Community/sources) instead.

Skittyblock:
closing this due to previous message; feel free to use the new api and put it somewhere other than this repo

PR #1035, opened 2025-09-26
Added new source "ru.mangabuff"

SolsticeLeaf:
## Changes
- New source

## Checklist:


src/rust/ru.mangabuff/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+opt-level = "s"
+strip = true
+lto = true
+
+[dependencies]
+aidoku = { git = "https://github.com/Aidoku/aidoku-rs", features = ["helpers"] }
+const_format = "0.2.31"
+itertools = { version = "0.10.5", default-features = false, features = [
+    "use_alloc",
+] }
```
Skittyblock, line 23:
```suggestion
] }

```

Skittyblock (dismissed):
ideally, more sources like this shouldn't be created with the old api, and they shouldn't be put here since the repo is meant to not be actively maintained anymore. but since it's already made i'll approve it

PR #1037, opened 2025-10-04
fix(mangabz): chapter parser logic

dyphire:
Checklist:

### DIFF
--- a/src/rust/zh.mangabz/res/source.json
+++ b/src/rust/zh.mangabz/res/source.json
@@ -3,8 +3,8 @@
 		"id": "zh.mangabz",
 		"lang": "zh",
 		"name": "Māngabz",
-		"version": 2,
-		"url": "https://mangabz.com/",
+		"version": 3,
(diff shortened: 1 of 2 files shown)


src/rust/zh.mangabz/res/source.json
```diff
@@ -3,7 +3,7 @@
 		"id": "zh.mangabz",
 		"lang": "zh",
 		"name": "Māngabz",
-		"version": 2,
+		"version": 3,
 		"url": "https://mangabz.com/",
```
Skittyblock, line 7: looks like this slipped in by accident before, urls shouldn't have trailing slashes
```suggestion
		"url": "https://mangabz.com",
```

Skittyblock (changes requested):

src/rust/zh.mangabz/res/source.json
dyphire: Done

Skittyblock (approved):

PR #1040, opened 2025-10-06
fix: update Desu source domain

ywerfer:
## Fix: Updated Desu Source Domain

**What changed**
- Updated domain in `README.md`, `source.json`, and `constants.rs`
- Adjusted parsing and URLs to match the new domain

**Why**
- The old domain `desu.work` is no longer valid
- The new domain `desu.city` is active and required for source to function

### DIFF
--- a/src/rust/ru.desu/res/source.json
+++ b/src/rust/ru.desu/res/source.json
@@ -4,7 +4,7 @@
 		"lang": "ru",
 		"name": "Desu",
 		"version": 2,
-		"url": "https://desu.work",
+		"url": "https://desu.city",
 		"nsfw": 2
 	}
 }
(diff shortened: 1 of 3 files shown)


src/rust/ru.desu/res/source.json
```diff
@@ -4,7 +4,7 @@
 		"lang": "ru",
 		"name": "Desu",
 		"version": 2,
```
Skittyblock: version should also be bumped so people can update
```suggestion
		"version": 3,
```

Skittyblock (changes requested):

ywerfer:
Yeah, u`re right

Skittyblock (approved):

PR #1053, opened 2026-01-05
fix: update broken sources

### LINKED ISSUE #983: Hentaicube url change
Please change address vietnam source: hentaicb.bar to hentaicube.xyz

kkantan:
these were some broken sources that had very simple fixes, so I thought we might as well have them instead of removing.

updated:
- en.resetscans
- en.lilymanga
- vi.hentaicube: closes #983
- fr.sushiscan: no version increment, since there's already a newer version [here](https://github.com/Aidoku-Community/sources)
- id.manhwalist: no version increment, since there's already a newer version [here](https://github.com/Aidoku-Community/sources)


Skittyblock (approved):
i suppose

